PR #12, opened 2015-01-05
Add analytics integration via SEGMENT. Requires JEKYLL_ENV=production|development -> Added plugin to support that

cristiandouce:


cristiandouce:
@jfromaniello :dancers:

PR #53, opened 2015-08-14
Two factor authentication for node.js apps blog post.

sebadoom:
Ready for comments.


sebadoom:
Full screen capture

woloski:
Feedback
* We should have an aside at the end showing how it works on Auth0 (dashboard with a switch :))
* We should mention all the things that you have to implement for this to be production ready:
   - have a way to bypass this if you lost/change your phone (recovery codes)
   - have a way to reset the MFA for an admin user
   - fallback to SMS if the user does not have a smart phone
   - send emails every time mfa is turned on or off for your app
   - checkbox to remember the MFA for X days

jgelsey:
Typo: s/enrolment/enrollment/g

_posts/2015-08-14-add-two-factor-authentication-to-your-node-apps-with-passport.markdown
```diff
@@ -0,0 +1,206 @@
...
+```
+
+As usual, passport will call this callback with the actual 'user' object as deserialized from the request. In this case, we are using a simple JSON-based filesystem storage. In production, you should never store passwords and secrets as plain text. The 'user' object contains a 'key' string that may be null. In case the key is null, we will consider TOTP disabled and allow logins with just a username and password. In case the key is present, we will require an additional input from the user before proceeding with the login. The key must be provided to passport as a bytearray, but for convenience (and because Google Authenticator requires so) we store it as a BASE32 encoded string. The 'period' is the amount of time a single TOTP code es valid from the moment it is requested to the moment it is entered.
+
+The enrolment step requires the user to be already logged-in (as this is necessary to enable of disable TOTP in the server for each specific account). If the user chooses to enable TOTP, we present the shared-secret as a QR code ready to scan and save in Google Authenticator:
+
+```javascript
+app.get('/totp-setup', 
+    isLoggedIn,
+    ensureTotp,
```
mgonto: What's this `ensureTopt`?

mgonto: I'd explain it in words here (I see the code is below)

mgonto:
I really like it :).

Agreed with @woloski. Add a section on "It's implemented, now what?" or something like that and mention all the steps to get it production ready. And then put an `Aside` that shows how easy it's to set it up with Auth0, with a link to the Multifactor section in the dashboard and a screenshot showing Mutifactor enabled

mgonto:
@vctrfrnndz style for the post header and icon look good? Can you help with tha?

mgonto:
@jgelsey after the blogpost is reviewed, we send it to wordy.com, so all typos will be fixed :).

sebadoom:
@jgelsey Thanks for checking, I wasn't sure about it either. I did check the spelling: http://dictionary.reference.com/browse/enrolment, but I'll use whatever is commonplace :) I used one "l" because google seemed to like it more that way: https://www.google.com.ar/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=define%20enrollment

sebadoom:
Now awaiting more feedback :)

sebadoom:
Can't post a full page screenshot because the extension is rendering the page really bad :( I'll see if another extension works better and post a capture if possible.

sebadoom:
Pay no mind to the image in the aside section, it's a GIF, so it gets corrupted when captured. Sigh.
Full page capture

mgonto:
All links to CDN should be `cdn.auth0.com` instead of `http://assets.auth0.com/`. They work!

mgonto:
Other than that, I think it's good to go :)

mgonto:
@sebadoom did you send it to wordy?

sebadoom:
All changes done.
adding two-factor authentication to your node js express apps with passport js and google authenticator

woloski:
screenshots @mgonto

woloski:
too much text in some areas imo, can we try to cut some?

mgonto:
For the MFA screenshot, use https://cdn.auth0.com/blog/grow-revenue/mfa.png.

sebadoom:
@mgonto The GIF shows that part of the process. Just to confirm, do you still want to change it for the one you linked?

@woloski Let me see what I can do

mgonto:
Yep, still to the image.

woloski:
I would be more subtle. After the conclusion. "Aside: How it works in Auth0"

woloski:
And use bullets to explain the value prop

* Turn on/off, no changes to your application
* Integrates with any TOTP app (Google Authenticator, Microsoft Authenticator, Authy, ....)
...

dschenkelman:
Should we wait until I update the mfa widget with the new look?

mgonto:
@dschenkelman I don't think so

sebadoom:
We can update the picture later, I suppose.

sebadoom:
Implemented suggestions.
adding two-factor authentication to your node js express apps with passport js and google authenticator

sebadoom:
Err, sorry about that, missed the button. Sigh.

woloski:
"First steps to production" should be "Is this production ready?"

sebadoom:
from theory to practice adding two-factor to node js

sebadoom:
Sent to wordy, awaiting results.

woloski:
from a high level looks much better

sebadoom:
Wordy changes implemented.
from theory to practice adding two-factor to node js 2

PR #54, opened 2015-08-14
Introduce metricslib

vctrfrnndz:
Intruduce new metrics lib 

For review @cristiandouce

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -39,14 +39,6 @@
     <link href="//cdn.auth0.com/styleguide/latest/index.min.css" rel="stylesheet">
 
     <link rel="stylesheet" href="{{ site.baseurl }}/css/index.css" />
-
-    <script type="text/javascript">
-      var segmentKey = {% if 'production' == site.env %} 'aaB6w2BsRwUcGbRSUY7X22FPAa1NL3BL' {% else %} 'WqErZyd56ob3pmMDnko55hNgFt8B4Zox' {% endif %};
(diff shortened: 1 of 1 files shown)


_layouts/default.html
```diff
@@ -138,8 +198,8 @@
         });
 
         window.login = function login() {
-          if (window.analytics) {
-            window.analytics.track('click:signin', getBasicInfo());
+          if (metricsLib) {
```
mgonto: Is the if needed? /cc @cristiandouce it should work anyway, right?

_layouts/default.html
```diff
@@ -286,6 +348,8 @@
               initialize();
 
               NProgress.done();
+
+              metricsLib.page();
```
mgonto: Does this run on first load as well?

vctrfrnndz: Yes. Also just updated PR with some fixes. Should be good to go with @cristiandouce 's blessing

_layouts/default.html
```diff
@@ -121,13 +172,13 @@
         };
 
         lock.on('signup ready', function () {
-          if (!window.analytics) return;
-          window.analytics.track('open:signup', getBasicInfo());
+          if (!metricsLib) return;
```
cristiandouce: always use `window.metricsLib` please

cristiandouce: or define a getter function like
```js
function getMetrics() {
  return window.metricsLib;
}
// later

var metrics = getMetrics();

if (!metrics) { return; };
metrics.something()
```

gnandretta:
@cristiandouce I've verified the following events show up in Segment's debugger:
- click:signin
- open:signin
- click:signup
- open:signup
- success:blog-signup
- pages (when visiting for the first time and changes via pjax)

mgonto:
Cool :). Thanks @gnandretta and good catch. Ready to merge @cristiandouce right?

cristiandouce:
All good! :+1:

PR #57, opened 2015-08-28
Added Falcor article

chenkie:
Images are being served from Dropbox right now but we'll move them over to the CDN asap.


_posts/2015-08-28-getting-started-with-falcor-from-setting-up-a-json-graph-to-implementing-search.markdown
```diff
@@ -0,0 +1,446 @@
...
+However, when we try to run this we see that the data looks exactly the same and there's no `location` key. That's because Falcor only allows us to pull primitive JavaScript values from our models, and not arrays or objects. Data within arrays or objects can tend to become arbitrarily large over time, so Falcor makes it illegal for us to access them directly with the intent that we give better thought to crafting our paths and only retrieve the data that we need. If we want to get access to the values of the keys on the `location` object within our data, we can pass another array with those paths.
+
+```js
+// app.js
+
+...
+
+// To get the values on the "location" object, we need to pass the paths for the
+// keys on that object
+.get(["events", {from: 0, to: 2}, ["name", "description", "location"],["city", "state"]])
```
ggoodman: I don't think this is quite right. What this `pathSet` is asking is find me the `city` and `state` for each of `name`, `description` and `location` of the first 3 `events`. I _think_ that the problem with this query is that it would result in a (not yet connected) server trying to resolve `name.city`, `name.state`, etc..

`Falcor.Model#get` is a variadic function accepting one or more `pathSet`s. So the correct request would be more like:

```js
.get(['events', {from: 0, to: 2}, ['name', 'description']], ['events', {from: 0, to: 2}, 'location', ['city', 'state']])
```

This was my interpretation anyway. Might be worth confirming this with your test server.

_posts/2015-08-28-getting-started-with-falcor-from-setting-up-a-json-graph-to-implementing-search.markdown
```diff
@@ -0,0 +1,446 @@
...
+
+```js
+// index.js
+
+...
+
+{
+    // The search route will match keys that match the names
+    // of our conferences
+    route: "events.byName[{keys}]['description']",
```
ggoodman: A couple thoughts about the search method:
1. The search method should return `$ref`erences to the underlying events. This means there is no need to tack `['description']` on the end.
2. The search method may better be implemented as a `call`. I think this because the set of events whose names match the search key will change over time but Falor will not reflect these changes by invalidating the path.

PR #63, opened 2015-09-04
Microservices logo served over https

chenkie:
Changed microservices logo to pull from https instead of http to avoid mixed content warning.


mgonto:
Nice catch!

PR #64, opened 2015-09-07
Added ABTesting library to blog

cabralmartin:
* Added ABTest library to blog

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -97,12 +97,16 @@
 
           executeABExperiments();
         } catch (e) {
-          null !== console && console.error('Error running AB experiments.');
+          console.error('Error running AB experiments.');
+          $('#pjax').css('visibility', 'visible');
+          window.abLoaded = true;
         }
       });
 
     } catch (e) {
-      null !== console && console.error('Error initializing ABTesting lib. ABTesting will be disabled');
+      console.error('Error initializing ABTesting lib. ABTesting will be disabled');
+      $('#pjax').css('visibility', 'visible');
+      window.abLoaded = true;
     }
     </script>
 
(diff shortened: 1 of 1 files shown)


_layouts/default.html
```diff
@@ -77,6 +79,74 @@
...
+    }
+    </script>
+
+    <script>
+    function executeABExperiments () {
+      var path = window.location.pathname;
+      var experiments = window.experiments;
+
+      if(!window.abLoaded) {
+        return;
```
cristiandouce: what would happen if we enter here?
Do we have some timeout... in case we never reach a any $visibility down this?

_layouts/default.html
```diff
@@ -77,6 +79,74 @@
...
+    <script src="//cdn.auth0.com/js/ab/ab-0.js"></script>
+    <script src="//cdn.auth0.com/js/ab/ab-loader-0.js"></script>
+
+    <script>
+    try {
+      window.abTestingLib.load({ fetchFn: function () { return $.get("https://ab.it.auth0.com/"); }});
+      window.abTestingLib.integration('auth0metrics').start({ metrics: window.metricsLib});
+      window.abTestingLib.start();
+
+      window.abTestingLib.onReady(function (ab) {
```
mgonto: What if ready is never called :)?

_layouts/default.html
mgonto: This is sync, but from what I understand, the "onReady" can be called after this script is executed since onReady is async. So this wouldn't work in lots of cases, right?

cabralmartin: The AB lib has a watchdog of 1500ms. When that timer expires the library fires the `error` event which will run the `onReady` handler and, in this case, will run the `executeABExperiments` function.

_layouts/default.html
cabralmartin: Same as above.

_layouts/default.html
cabralmartin: I could add a new watchdog on the blog's code in order to be extra sure that the content will be displayed if that's your preference. Thoughts?

PR #77, opened 2015-09-16
Angular 2 Models article

chenkie:
screen-capture

### DIFF
--- a/_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
+++ b/_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
@@ -2,7 +2,7 @@
 layout: post
 title: "Angular 2 Series - Part 2: Domain Models and Dependency Injection"
 description: "Learn about Angular 2's new Dependency Injection system and how to use models to organize your app."
-date: 2015-09-16 11:00
+date: 2015-09-17 11:00
 author: 
   name: Ryan Chenkie
   url: https://twitter.com/ryanchenkie?lang=en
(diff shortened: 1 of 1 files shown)


_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+    <input type="checkbox" [(ng-model)]="user.php">PHP
+    <button>Submit</button>
+  </form>
+```
+
+## Handling User Data
+
+For our app, we'll want to take the user's input and see how it can be handled with a model. For simplicity, we'll simply log the info to the console in these examples, but you would typically be doing HTTP requests to send the user data off to a server to be persisted.
+
+We could set up a simple `User` class that acts as a model for our data and have some additional methods in our `UsersAppComponent` that act upon the form data.
```
mgonto: Let's mention problem and drawabcks of this case here and why we need the next one.

Let's also add a graphic showing that the code for the method is actually in the Component not in the User, something like a Class Diagram

mgonto: I see explanation in the next section, maybe move it to this section to then go to Improving. What do you think?

_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+
+## Improving the Model
+
+While this approach works sufficiently when an application is small, the way we have set things up does have a few drawbacks:
+
+1. The `User` class isn't enforcing which values should be present or not. This can have implications if some operation downstream expects certain values to always be present.
+2. The model doesn't have methods of its own to handle calculation of the user's rating or submitting the data. What if we wanted to have this same functionality in another of our app's components? We would need to duplicate the code in those other components.
+3. While `calculateRating` is trivial in this example, what if we had some important and sensitive method with business logic that is meant only for certain circumstances? We need a way to make sure such methods can't be used in places they're not supposed to be.
```
mgonto: Ability to share models with Backend if it's JS ;)

_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+
+export class User {
+  name: string;
+  email: string;
+  rating: number;
+
+  constructor(userInfo:any) {
+    this.name = userInfo.name;
+    this.email = userInfo.email;
+    this.rating = this.calculateRating(userInfo);
```
mgonto: Show option of calculating it every time.

_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  private calculateRating(userInfo) {
+    var rating = 0;
+    userInfo.javascript ? rating+= 30 : rating;
```
mgonto: This ternary operator with side effect is kind of weird. Do a regular if or `userInfo.javascript && rating+=...`

_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+...
+
+class UsersAppComponent {
+  userName: string;
+  userRating: number;
+  rating: number;
+  
+  // We inject the UserFactory
+  constructor(@Inject(UserFactory) UserFactory) {  
+    this.user = UserFactory;
```
mgonto: I'd set this.UseFactory here

_posts/2015-09-15-angular-2-series-part-2-domain-models-and-dependency-injection.markdown
```diff
@@ -0,0 +1,371 @@
...
+bootstrap(UsersAppComponent, [UserFactory]);
+
+```
+
+With `this.user` pointing to the `UserFactory`, we can now make use of its methods. We can do this in our right in our component class, but we can also use it directly in the view.
+
+```html
+  <!-- usersTemplate.html -->
+
+  <form role="form" #form="form" (ng-submit)="user.create(form.value).save()">
```
mgonto: I don't like calling it directly from here, you can't handle success and error, do it from JS

mgonto:
There's a really important part missing which is about using models that already exist.

How can you use a Model that already exists? I'd show how to add a static `User.Get` method to the class or factory to show how to get an existing User in a particular page instead of just saving

chenkie:
SEO checklist: https://docs.google.com/a/auth10.com/document/d/1-W6BVt1EaNrCMA-IUHSvwQUFBlk7XRgkhQyhxJyYh-o/edit?usp=sharing

PR #86, opened 2015-09-23
Add shiny-server post

sicarul:
New blog post for shiny-server!


sicarul:
@sebadoom good point, i have no idea, i just wanted to include the possibility to signup directly there...

@mgonto can you tell me if it's safe to use? Thanks!

PR #101, opened 2015-10-05
Refresh tokens draft

sebadoom:
Refresh tokens article draft. Complete text. Missing diagrams, example and logo.
refresh tokens when to use them and how they interact with jwts


jaredhanson:
I know this is what we currently implement, but our use of `/delegation` to exchange a refresh token for a new id token is ill-advised and not to spec.

Refresh tokens are specified part of OAuth 2.0, and should be sent to `/token`, with a grant type of `refresh_token`.  When used in the context of OpenID Connect, refreshing id_tokens is described in this section of the spec: http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens

sebadoom:
Added draft diagrams
refresh tokens when to use them and how they interact with jwts

sebadoom:
Added example (please review). Still missing: redesigned diagrams and icon.
refresh tokens when to use them and how they interact with jwts

PR #131, opened 2015-11-08
Fix typo

HeroicEric:
"does this token **has** the right scope" -> "does this token **have** the right scope"


mgonto:
Thanks!!!

PR #141, opened 2015-11-20
Fixed Relay Article URL

plaurino:


mgonto:
You deleted the Relay article, that's not fixing it :D

PR #145, opened 2015-11-29
How to Implement Slack-like Login on iOS with Auth0

sebadoom:
How to Implement Slack-like Login on iOS with Auth0 draft. Full post. Bug in example. Some images are videos (the ones with the OS X dock visible). RFC.
how to implement slack-like login on ios with auth0

### DIFF
--- a/_posts/2015-11-29-how-to-implement-slack-like-login-on-ios-with-auth0.markdown
+++ b/_posts/2015-11-29-how-to-implement-slack-like-login-on-ios-with-auth0.markdown
@@ -35,28 +35,29 @@ In this post we will explore how to implement a Slack-like login strategy (email
 -----
 
 ## Introduction
-Passwordless login systems are convenient for end users: just remember your username, phone number or email and off you go. Passwordless systems usually require the user to manually input a code to validate their identity. This is no longer the case thanks to the advent of [Universal Links](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/AppSearch/UniversalLinks.html). Once an email with a link is received by the user, he or she can simply tap on the link and have the application take over. And the cool things is: it is that easy.
+Passwordless login systems are convenient for end users: just remember your username, phone number or email and off you go. Passwordless systems usually require the user to manually input a code to validate their identity. This is no longer the case thanks to the advent of [Universal Links](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/AppSearch/UniversalLinks.html). Once an **email with a link** is received by the user, he or she can simply **tap on the link and have the application take over**. And the cool things is: it is that easy.
 
 ![Slack-like login](https://cdn.auth0.com/blog/iosmagiclink/slack_magiclink.jpg)
 
 ## Universal Links
 Apple has long provided a way to *connect* applications through URLs: [URL schemes](https://developer.apple.com/library/ios/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/Inter-AppCommunication/Inter-AppCommunication.html#//apple_ref/doc/uid/TP40007072-CH6-SW1). URL schemes have their own set of issues such as [no fallback mechanism](http://stackoverflow.com/questions/6964515/launching-app-or-app-store-from-safari/6965646#6965646) and [URL scheme hijacking](http://stackoverflow.com/questions/33919058/prevent-ios-url-scheme-hijack). This has pushed Apple to develop a new and safer way to tell iOS when a link should be passed to an app installed on the device. This is known as [Universal Links](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/AppSearch/UniversalLinks.html).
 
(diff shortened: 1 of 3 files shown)


sebadoom:
Use applinks instead of webcredentials and add an additional warning for the use of Gmail.
how to implement slack-like login on ios with auth0 1

woloski:
would be great a video of the whole ux

hzalaz:
@sebadoom @woloski we should not provide any "ready to use" sample since it wouldn't work since they will need access to our apple dev account. This is because we specified that apps with our Apple Dev TeamID should open those links and apple validates that with the certs you use to sign the app before running it.
My suggestion will be to use the passwordless sample and tell them to set the flag in Lock and add the entitlements in Xcode, the rest is just account configuration.

hzalaz:
@sebadoom you can use 
```swift
A0Lock.sharedLock()
```
to get a shared lock instance that loads Auth0 credentials from `Info.plist` file.

Also this will only work in iOS 9, Lock checks that and falls back to code input if running in iOS 8 or lower

sebadoom:
@hzalaz I agree. I will delete the sample repo and tell people in the post to use the generic samples as base. I will also add a note clarifying magic link only works on iOS 9 and Lock automagically fallsback to code input on previous versions.

@woloski let me see if I can take a full video capture of the login process. I will need to set up a different email than the one I normally use since the Gmail app opens links internally. Where would you like me to put it in the post? At the end?

sebadoom:
Latest draft, switched repos to the mobile samples one, added full flow video. Missing: icon
how to implement slack-like login on ios with auth0 2

hzalaz:
@sebadoom I'd rather use the link to download the seed sample (configured with their account credentials if they are logged in) instead of making them clone the repo with all the samples and navigate a directory.

sebadoom:
@hzalaz Oh, where's that? Let me see if I can find it.

hzalaz:
@sebadoom it's not documented but we use it in our docs site, not sure if it'll work in the blog though. Check the get seed project button [here](https://auth0.com/docs/connections/passwordless/ios-email-swift)

sebadoom:
@hzalaz I don't think that will work. The client id is set in the URL: https://auth0.com/docs/Mobile-Samples.iOS/master/create-package?path=Passwordless-Email/Lock/Swift&filePath=Passwordless-Email/Lock/Swift/Passwordless-Email/Info.plist&type=replace&clientId=FyFnhDX2kSqtpMZ6pGe6QpQuJmD7s4dj

hzalaz:
@sebadoom without that parameter it should work too, otherwise just add a link to the directory in the repo of samples.

sebadoom:
@hzalaz Just tested the link without the client id. It creates a ZIP file without changing the Auth0 ID, so it's just like cloning it from GitHub.

hzalaz:
@sebadoom some last comments, we should use the domain `samples.auth0.com`, the app screenshot should have the device frame and finally we should use 

```swift 
A0Lock.sharedLock()
```

instead of the singleton app, I need to change the samples to use that but it's the same behaviour but does not require the user to keep a singleton.

Also for the request, I'd rather put the curl command, and maybe output rather than the screenshot to our site that does not give any help (we can add the link).

sebadoom:
We should use samples.auth0.com for what? As an Auth0 test account?

I didn't use A0Lock.sharedLock because of the way the samples create the instance of the A0Lock object (they don't use sharedLock, and by taking a look at the Lock code I noticed sharedLock uses a different instance from the ones returned by other constructors). So until the samples are updated we cannot tell people to do that based on our example.

I did not use the Curl command because then we need to explain how to get the JWT, and that means people need to browse to that page anyway.

hzalaz:
`samples.auth0.com` is to avoid adding your domain `speyrott.auth0.com`.
About `A0Lock.sharedLock()`, the samples are coded that way since I don't have time to keep everything up to date and build stuff so they are a little outdated.

Good point about the JWT, it's harder to explain how to get one.

hzalaz:
A little more for `A0Lock.sharedLock()` is that we should use it since I want this code snippet to be easily use by copy-pasting without adding a lot of dependencies like creating that `Application` class, I'd make these snippets agnostic of the sample. e.g.: I'd replace 

```swift
controller.onAuthenticationBlock = { (profile, token) in
    let app = Application.sharedInstance
    app.token = token
    app.profile = profile
    self.dismissViewControllerAnimated(true, completion: { self.performSegueWithIdentifier("UserLoggedIn", sender: self) })
}
```

and use 

```swift
controller.onAuthenticationBlock = { (profile, token) in
   // You're authenticated
    self.dismissViewControllerAnimated(true, completion:nil)
}
```

sebadoom:
@hzalaz I undestand. Would you like me to send a PR for the iOS samples with the sharedLock change? My point is we need to have this available when the post goes live (ASAP) as people are bound to follow the steps and try it.

sebadoom:
@hzalaz it would also include updates pods.

PR #188, opened 2016-01-21
JavaScript Leaks Article

sebadoom:
Request for comments. We are still missing the part where we actually find memory leaks and some parts where we explain the profiling tools.

four types of leaks in your javascript code and how to get rid of them


sebadoom:
Complete article. RFC. Icon needed.
four types of leaks in your javascript code and how to get rid of them 1

_posts/2016-01-26-four-types-of-leaks-in-your-javascript-code-and-how-to-get-rid-of-them.markdown
```diff
@@ -0,0 +1,304 @@
...
+
+-----
+
+## Introduction
+Memory leaks are a problem every developer has to face eventually. Even when working with memory-managed languages there are cases where memory can be leaked. Leaks are the cause of whole class of problems: slowdowns, crashes, high latency, and even problems with other applications.
+
+### What are memory leaks?
+In essence, memory leaks can be defined as memory that is not required by an application anymore that for some reason is not returned to the operating system or the pool of free memory. Programming languages favor different ways of managing memory. These ways may reduce the chance of leaking memory. However, whether a certain piece of memory is unused or not is actually an [undecidable problem](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#Release_when_the_memory_is_not_needed_anymore). In other words, only developers can make it clear whether a piece of memory can be returned to the operating system or not. Certain programming languages provide features that help developers do this. Others expect developers to be completely explicit about when a piece of memory is unused.
```
chenkie: Anything we can link to here for reading more about different ways of managing memory?

_posts/2016-01-26-four-types-of-leaks-in-your-javascript-code-and-how-to-get-rid-of-them.markdown
```diff
@@ -0,0 +1,304 @@
...
+
+// Foo called on its own, this points to the global object (window)
+// rather than being undefined.
+foo();
+```
+
+> To prevent these mistakes from happening, add `'use strict';` at the beginning of your JavaScript files. This enables a stricter mode of parsing JavaScript that prevents accidental globals.
+
+#### A note on global variables
+Even though we talk about unsuspected globals, it is still the case that much code is littered with explicit global variables. These are by definition noncollectable (unless nulled or reassigned). In particular, global variables used to temporarily store and process big amounts of information are of concern. If you must use a global variable to store lots of data, make sure to null it or reassign it after you are done with it. One common cause for increased memory consumption in connection with globals are *caches*. Caches store data that is repeatedly used. For this to be efficient, caches must have an upper bound for its size. Caches that grow unbounded can result in high memory consumption because their contents cannot be collected.
```
chenkie: Anything we can link to for more info on caches?

_posts/2016-01-26-four-types-of-leaks-in-your-javascript-code-and-how-to-get-rid-of-them.markdown
```diff
@@ -0,0 +1,304 @@
...
+- [JScript Memory Leaks - Douglas Crockford (old, in relation to Internet Explorer 6 leaks)](http://javascript.crockford.com/memory/leak.html)
+- [JavaScript Memory Profiling - Chrome Developer Docs](https://developer.chrome.com/devtools/docs/javascript-memory-profiling)
+- [Memory Diagnosis - Google Developers](https://developers.google.com/web/tools/chrome-devtools/profile/memory-problems/memory-diagnosis)
+- [An Interesting Kind of JavaScript Memory Leak - Meteor blog](http://info.meteor.com/blog/an-interesting-kind-of-javascript-memory-leak)
+- [Grokking V8 closures](http://mrale.ph/blog/2012/09/23/grokking-v8-closures-for-fun.html)
+
+## Conclusion
+Memory leaks can and do happen in garbage collected languages such as JavaScript. These can go unnoticed for some time, and eventually they can wreak havoc. Memory profiling tools are essential in finding memory leaks. Profiling runs are essential for mid or big-sized applications and should be part of the development cycle, don't be afraid to start doing this if you want the best possible user experience. Hack on!
```
diegopoza: I like the conclusion but I think we can make it stronger. Also you are using the word "essential" twice in contiguous sentences. Below is my proposed changes, feel free to use it or edit it at will.

Proposed conclusion: "Memory leaks can and DO happen in garbage collected languages such as JavaScript. These can go unnoticed for some time, and eventually they will wreak havoc. For this reason, memory profiling tools are essential for finding memory leaks. Profiling runs should be part of the development cycle, especially for mid or big-sized applications. Start doing this to give your users the best possible experience. Hack on!"

PR #189, opened 2016-01-22
Angular 2 Routing Post

chenkie:
Screen capture didn't work (too large).
SEO Checklist: https://docs.google.com/a/auth10.com/document/d/1sz6-NOuIPnfnjqEHHl6BluBpdSe9lbhzk8EFrYG0E4k/edit?usp=sharing

### DIFF
--- a/_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
+++ b/_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
@@ -28,6 +28,8 @@ tags:
 
 **TL;DR:** The new component router in Angular 2 comes with a lot of features and a slightly different approach to setting up routing. In this article we take a look at the router's features by implementing them in a simple Github user explorer application. Follow along here or take a look at [the repo](https://github.com/auth0/angular2-routing) to get started with the code.
 
+If you'd like to see our other Angular 2 content, we've also covered [pipes](https://auth0.com/blog/2015/09/03/angular2-series-working-with-pipes/), [dependency injection](https://auth0.com/blog/2015/09/17/angular-2-series-part-2-domain-models-and-dependency-injection/), [HTTP](https://auth0.com/blog/2015/10/15/angular-2-series-part-3-using-http/), and [authentication](https://auth0.com/blog/2015/11/10/introducing-angular2-jwt-a-library-for-angular2-authentication/).
+
 ---
 
 Angular 2 comes with a brand new component router that is packed with features. At the Beta 1 release, the documentation on Angular 2's router is in pretty good shape, and there is even an [extensive guide](https://angular.io/docs/ts/latest/guide/router.html) on setting it up. There are, howevever, some features and use cases that have yet to be documented.
@@ -38,7 +40,7 @@ In this article we'll get a feel for the new router by implementing it in a simp
 
 ![angular2 routing router](http://cdn.auth0.com/blog/angular2-routing/angular2-routing-1.png)
 
-## Getting Started
+## Getting Started With the Router
 
 It's best to get started with a seed project. If you'd like to use Webpack, the [Angular 2 Webpack Starter](https://github.com/AngularClass/angular2-webpack-starter) from [AngularClass](https://angularclass.com/) is an excellent option.
 
--- a/_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
+++ b/_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
@@ -1,0 +1,658 @@
+---
+layout: post
+title: "Angular 2 Series - Part 4: Component Router In-Depth"
+description: "Learn all about Angular 2's new component router and how to set up routing in your app."
+date: 2016-01-25 08:30
+author: 
+  name: Ryan Chenkie
+  url: https://twitter.com/ryanchenkie?lang=en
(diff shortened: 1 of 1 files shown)


_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+
+In this article we'll get a feel for the new router by implementing it in a simple Github user explorer application. In the app, we'll be able to search for Github accounts based on username, and we'll set up a way to explore further detail about the users. To do this, we'll set up child routes and work with route parameters.
+
+![angular2 routing router](http://cdn.auth0.com/blog/angular2-routing/angular2-routing-1.png)
+
+## Getting Started With the Router
+
+It's best to get started with a seed project. If you'd like to use Webpack, the [Angular 2 Webpack Starter](https://github.com/AngularClass/angular2-webpack-starter) from [AngularClass](https://angularclass.com/) is an excellent option.
+
+The router in Angular 2 provides both a `HashLocationStrategy` and a `PathLocationStrategry`, one of which is used when we bootstrap the app. As you might guess, `HashLocationStrategy` uses hashes in the URL, which is what we're used to as the default from Angular 1.x. Many prefer not to use hashes, in which case `PathLocationStrategy` can be applied.
```
diegopoza: Would love to see sample URLs using both strategies. Maybe in a note or an image showing the browser's address bar for both cases?

_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+import {Component} from 'angular2/core';
+
+@Component({
+  template: `<h1>Search for a Github user and view their profile</h1>`
+})
+
+export class Home {}
+```
+
+Now we need to set up our initial routing configuration. This is done with the `@RouteConfig` decorator, which takes an array of objects called `RouteDefinitions` that describe our app's various routing paths and their respective components. We already imported `RouteConfig`, so let's now set it up to just have the `App` class definition.
```
sebadoom: Even though it is obvious from the example below, I'd be explicit about `@RouteConfig` being a decorator for the App class (as opposed to the component definition classes).

_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+import {ROUTER_DIRECTIVES, RouteParams, RouteConfig} from 'angular2/router';
+
+@Component({
+  template: `
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <h1>{{userLogin}}</h1>
+      </div>
+      <div class="panel-body">
+        <router-outlet></router-outlet>
```
sebadoom: It might be worth pointing out what this instance of `router-outlet` will do next to this code (i.e. in the paragraph below). I guess this will show additional user information that another component will provide. This will probably be explained somewhere below. I can picture someone getting confused by the nested uses of `router-outlet` otherwise.

_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+@RouteConfig([
+  { path: '/detail', component: UserDetail, name: 'UserDetail' }
+])
+
+...
+```
+
+`User` has now become a **routing** component instead of stictly a **view** component because it is facilitating further routing. In fact, this is by design in Angular 2's router, but can be undesirable because it causes the need for additional files. The router's design means that the ideal routing setup should look like this:
+
+![angular2 routing router](http://cdn.auth0.com/blog/angular2-routing/angular2-routing-diagram.png)
```
sebadoom: This diagram is really helpful. Good job deciding to put it.

_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+      .subscribe(
+        data => this.userData = data,
+        err => console.log(err)
+      );
+  }
+  
+}
+```
+
+You'll notice that the way we're getting the `userLogin` param is quite different in this case. We can't get at the params of the parent route by simply using `params.get` like we previously did, and instead we need to use the `Injector` to reference the `RouteParams` of the parent. For nesting that doesn't go too deep this isn't a big deal, but as we'll see later, it can get cumbersome when we nest further.
```
sebadoom: Any reasons as to why this was designed this way?

_posts/2016-01-25-angular-2-series-part-4-component-router-in-depth.markdown
```diff
@@ -0,0 +1,660 @@
...
+
+There's a problem with the route configuration above, and we can take this opportunity to demonstrate the non-terminal route error. If we run it and try to select a user, we can see the error in the console.
+
+![angular2 routing router](http://cdn.auth0.com/blog/angular2-routing/angular2-routing-5.png)
+
+The issue is that we need to indicate that the `/detail` route is the default one to use, and thus make it the terminal route.
```
sebadoom: Why is this the case? Isn't `/detail` a route that will get set as a path explicitly when clicking? Is it because a default route is always necessary?

PR #219, opened 2016-03-07
Added HapiJS auth article

chenkie:


_posts/2016-03-07-hapijs-authentication-secure-your-api-with-json-web-tokens.markdown
```diff
@@ -0,0 +1,586 @@
...
+    },
+    // Validate the payload against the Joi schema
+    validate: {
+      payload: createUserSchema
+    }
+  }  
+}
+```
+
+Hapi routes need to have a route `method` and `path` at a minimun, and a `handler` if they are to be useful. Configuring these details here is self-explanatory, but a few things are likely unfamiliar. Down at the bottom, we have a spot for validating input, and in this case, we want to validate the `payload` that comes in. If we were accepting params from the user, then we could specify our validation on the `params` key. This validation is coming from the `createUserSchema` within the `schemas` subdirectory.
```
sebadoom: s/minimun/minimum

_posts/2016-03-07-hapijs-authentication-secure-your-api-with-json-web-tokens.markdown
```diff
@@ -0,0 +1,586 @@
...
+}
+```
+
+We've specified that this route should implement the `jwt` auth strategy (which we defined in `server.js`), and that the user must have a scope of `admin` to access the route. If we [inspect our JWT](http://jwt.io), we can see that we've got a `scope` of `admin`. 
+
+![hapijs authentication](https://cdn.auth0.com/blog/hapi/hapi-auth-3.png)
+
+You might be wondering if this is secure. Since we can inspect and change the content of the JWT in the debugger, couldn't a malicious user either change an existing JWT or create a new one that could compromise the API? Remember that the beauty of JWTs is that they are digitally signed with the secret key on our server. For the modified JWT to be valid, the attacker would need to know the secret. As long as we have a strong private key, our JWTs are secure.
+
+![hapijs authentication](https://cdn.auth0.com/blog/hapi/hapi-auth-4.png)
```
sebadoom: Heads-up: personal mail shown in picture.

PR #305, opened 2016-05-20
Add post about Sumo Logic extension

eugk:


eugk:
Needs a new date, short description, probably different tags, etc.

eugk:
Do not merge, I need to make updates.

PR #309, opened 2016-05-25
Finished JWTs in Ruby post

searsaw:


_posts/2016-05-25-jwts-in-ruby.markdown
```diff
@@ -0,0 +1,343 @@
...
+      [403, { 'Content-Type' => 'text/plain' }, ['The token does not have a valid issuer.']]
+    rescue JWT::InvalidIatError
+      [403, { 'Content-Type' => 'text/plain' }, ['The token does not have a valid "issued at" time.']]
+    end
+  end
+
+end
+```
+
+This is a simple Rack middleware setup. In the `call` method, we get the `HTTP_AUTHORIZATION` value out of the environment hash, which represents the `Authorization` header in the request.  We use `fetch` so we get a string if `HTTP_AUTHORIZATION` doesn't exist and then use `slice` to prevent throwing an error if the string indexes we want don't exist. This will strip off the "Bearer: " part of the `Authoriation` header. We then decode the JWT and add the scopes and user data to the environment hash. Lastly, we call the next part in our middleware stack and pass it our modified environment.
```
diegopoza: Typo in "part of the `Authoriation` header" -> Authorization.

_posts/2016-05-25-jwts-in-ruby.markdown
```diff
@@ -0,0 +1,343 @@
...
+
+```
+source "https://rubygems.org"
+
+gem 'sinatra'
+gem 'jwt'
+gem 'json'
+```
+
+Then, run `bundle`. If you don't already have bundler installed, run `gem install bundler`. This will install the dependencies we need. Next we need to create our `config.ru` to run our Sinatra application.  Put the following in a file called `config.ru`.
```
diegopoza: Our `config.ru` **file**

_posts/2016-05-25-jwts-in-ruby.markdown
```diff
@@ -0,0 +1,343 @@
...
+  end
+
+end
+```
+
+This code sets up our two application routes. The only public route we have right now is `/login`. It authenticates a user based on a username and password. Currently, it just returns a confirmation message. This is where we will eventually return a JWT the user can use to authenticate with the API. The API only has one route for now also called `/money`. It returns a simple message now too. Notice, we have hardcoded the values for our users. Normally this would come from a database of some sort.
+
+Run `rackup` on the command line to start the server. Here are what our responses look like if we hit each of these endpoints.
+
+![First Successful Hit](https://cdn.auth0.com/blog/ruby-auth/first_successful_hit.png)
```
diegopoza: If you are going to put several contiguous images, please add captions or else it looks weird.

PR #436, opened 2016-09-17
Rxjs weather app

searsaw:


_posts/2016-09-17-building-a-weather-application-using-rxjs.markdown
```diff
@@ -0,0 +1,267 @@
...
+// Get stream of button clicks
+const btnClickStream =
+  Rx.Observable
+    .fromEvent(addLocationBtn, 'click')
+    .map(() => true)
+    .forEach(val => console.log('btnClickStream val', val));
+```
+
+Here is our first look at RxJS! We are using the `fromEvent` method on the `Rx.Observable` object to create a stream from the click event that will be emitted by our `addLocationBtn`. This means, any time it is clicked, that event object will be sent down this `btnClickStream`. Because we don't really care about any of the information in that event, we will use the `map` method on the returned stream to map each value to the value `true`. This just makes things easier to think about. Lastly, to make sure it's working, we use `forEach`, which adds a subscriber to the stream. Here we are simply logging the value.
+
```
sebadoom: Is using `map` for this standard practice? If so, consider saying so here.

_posts/2016-09-17-building-a-weather-application-using-rxjs.markdown
```diff
@@ -0,0 +1,267 @@
...
+The second function also takes a zip code. We return a stream by using the `fromPromise` method on the `Rx.Observable` object to create one from the promise that is returned from calling our `getTemperature` function and passing it the current zip code. Since this returns a stream, we can operate on it with any of the available methods that `Rx.Observable` instances have on them. Since we only care about the current zip code and its temperature, let's return an object that only has these pieces of data on them. We `map` over the promise stream, destructuring the input to get the temperature out, and return an object containing the data we want.
```
sebadoom: Consider putting a link to ES2015 destructuring here.

_posts/2016-09-17-building-a-weather-application-using-rxjs.markdown
```diff
@@ -0,0 +1,267 @@
...
+    locationEle.appendChild(tempEle);
+    appContainer.appendChild(locationEle);
+
+    zipcodeInput.value = '';
+  });
+```
+
+We take the stream of zip codes and call `flatMap` on it. `flatMap` works just like regular `map` except it takes a stream of streams, pulls the value out of each stream, and makes it available in the stream returned from the method. This means it will flatten out the streams created by our request to the weather API and give us the values from those streams to act on, namely, the objects of zip codes and temperatures.
+
+The last part uses our trusty `forEach` to get each piece of data and add it to the page. Everything in the callback is vanilla JavaScript. We construct the elements and add them to the page. Then we empty out the value of the zipcodeInput.
```
sebadoom: Missing code tag here (zipcodeInput).

PR #626, opened 2017-02-07
Glossary of Modern JavaScript Concepts - Part 1

kmaida:
Please review, thanks!


_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
+---
+layout: post
+title: "Glossary of Modern JavaScript Concepts: Part 1"
+description: "Learn the fundamentals of functional programming, reactive programming, and functional reactive programming in JavaScript."
+date: 2017-02-09 8:30
+category: Glossary, Technical guide
+banner:
+  text: "Auth0 makes it easy to add authentication to your JS application."
```
brunokrebs: What exactly this `banner` option does? Besides that I couldn't understand how this sentence is related to the content itself.

kmaida: It's the Auth0 promo at the bottom of every blog post. This way, we can specify the copy for the banner promo to tailor it a little bit to be more relevant to the article. Ie., I said "to your JS application", whereas if the post had been about, say, PHP, I would customize the banner accordingly to say "your PHP application".

screen shot 2017-02-07 at 1 12 47 pm

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+
+Let's briefly discuss this implementation step-by-step.
+
+First, we'll create an <a href="#observables" target="_self">observable</a> called `time$`. This is a timer that adds a value to the collection every `1000ms` (every second). We need to `map` the timer event to extract its `value` and push it in the resulting stream.
+
+Next, we'll create a `move$` observable from the `document.mousemove` event. Mouse movement is _continuous_. At any point in the sequence, there are an infinite number of points in between. We'll throttle this so the resulting stream is more manageable. Then we can `map` the event to return an object with `x` and `y` values to represent mouse coordinates.
+
+Next we want to merge the `time$` and `move$` streams. This is a _transformation_. This way we can plot which mouse movements occurred during each time interval. We'll call the resulting observable `source$`. We'll also limit the `source$` observable so that it completes after ten seconds (`10000ms`).
```
brunokrebs: Merge is considered a transformation? Since it was the first time you addressed this term I decided to check it and what I got are [links like this](https://github.com/ReactiveX/RxJava/wiki/Transforming-Observables). No mention to `merge` over there.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+function sayHi() {
+  alert('Hi!');
+}
+function greet(greeting) {
+  greeting();
+}
+greet(sayHi); // alerts "Hi!"
+```
+
+> **Note:** When _passing a named function as an argument_, as in the two examples above, we don't use parentheses `()`. This way we're passing the function as an object. Parentheses _execute_ the function and return the result.
```
kmaida: 👍 A great suggestion. Making this change.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+
+Higher-order functions can also return another function:
+
+```js
```
kmaida: 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+* [Introducing the Observable](https://egghead.io/lessons/javascript-introducing-the-observable)
+* [RxMarbles](http://rxmarbles.com/)
+* [Rx Book - Observable](https://xgrommx.github.io/rx-book/content/observable/index.html)
+* [Introducing the Observable](https://egghead.io/lessons/javascript-introducing-the-observable)
+
+---
+
+## <span id="reactive-programming"></span>Reactive Programming
+
+**Reactive programming** is concerned with propagating and responding to incoming events over time, <a href="#imperative-delcarative" target="_self">declaratively</a> (describing _what_ to do rather than _how_).
```
kmaida: 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
kmaida: :thumbsup: You're absolutely correct, merge is technically not a transformation, it's a combining operator. Had map/filter on the mind when I wrote that-- great catch!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+* [Immutable Objects with Object Freeze](http://adripofjavascript.com/blog/drips/immutable-objects-with-object-freeze.html)
+* [Mutable vs Immutable Objects](https://www.interviewcake.com/concept/java/mutable)
+* [Using Immutable Data Stuctures in JavaScript](http://jlongster.com/Using-Immutable-Data-Structures-in-JavaScript)
+* [Getting Started with Redux](https://egghead.io/courses/getting-started-with-redux) (includes examples for addressing immutable state)
+
+---
+
+## <span id="imperative-declarative"></span>Imperative and Declarative Programming
+
+While some languages are **imperative** (C#, Java, PHP) or **declarative** (SQL, HTML), JavaScript supports both paradigms.
```
kmaida: Interesting. I think the catch here is languages that were "designed to primarily support" imperative programming. More languages these days support both paradigms, so it's becoming slightly trickier to cite purely imperative languages. I'll do some more research, or at least trim the examples and add a qualifier.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+var number = 1;
+function increment() {
+  return number++;
+}
+increment(); // global variable modified: number = 2
+```
+
+### Stateless
+
+**Stateless** programs, apps, or components perform tasks as though running them for the first time, every time. This means they do not reference or utilize any information from earlier in their execution. Statelessness enables _referential transparency_. Functions depend only on their arguments and do not access or need knowledge of anything outside their scope. <a href="#purity" target="_self">Pure functions</a> are stateless. See the following example:
```
kmaida: What I'm describing here is the definition of "stateless" as it applies directly to functional programming. Maybe it would help if I spelled that out a little more clearly, or added another paragraph addressing what you're referring to. I'll work on this a bit more, thanks!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,732 @@
...
+```
+
+In summary:
+
+* In pure functions, the same input will always produce the same output.
+* Pure functions rely only on local state and do not mutate external state.
+* Pure functions do not produce [side effects](https://en.wikipedia.org/wiki/Side_effect_%28computer_science%29).
+
+### Impure Functions
+
```
kmaida: :thumbsup: I'll add something about `random`, thanks!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+---
+
+## <span id="imperative-declarative"></span>Imperative and Declarative Programming
+
+While some languages were designed to be **imperative** (C++, PHP) or **declarative** (SQL, HTML), JavaScript (and others like [Java](http://openjdk.java.net/projects/lambda/) and [C#](https://msdn.microsoft.com/en-us/library/bb534803(v=vs.110).aspx)) can support both programming paradigms. 
+
+Most developers familiar with even the most basic JavaScript have written imperative code: instructions informing the computer _how_ to achieve a desired result. If you've written a `for` loop, you've written imperative JS. 
+
+Declarative code tells the computer _what_ you want to achieve rather than how, and the computer takes care of how to achieve the end result without explicit description from the developer. If you've used `Array.map`, you've written declarative JS.
```
sebadoom: Hmm, a simple SQL query seems a better example for this than `map`. In fact, IMHO `map` is a better example of functional aspects than declarative ones.

sebadoom: I see now that more examples like this follow. It is important to remark that declarative programming is usually concerned about describing transformations to data (and combining them) rather than specifying step-by-step recipes to achieve those same transformations. SQL embodies this very well, so as a first example it can do. You may want to stick with JS for later sections where you can go deeper into the meaning of declarative programming.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+> —_Conal Elliot_
+
+Functional reactive programming should be:
+
+* dynamic: can react over time _or_ to input changes
+* time-varying: reactive _behaviors_ can change continually while reactive _values_ change discretely
+* efficient: minimize amount of processing necessary when inputs change
+* historically aware: pure functions map state from a previous point in time to the next point in time; state changes concern the local element and not the global program state
+
+Conal Elliot's slides on the [Essence and Origins of FRP can be viewed here](http://conal.net/talks/essence-and-origins-of-frp-lambdajam-2015.pdf). An example of a functional reactive language is [Haskell](https://wiki.haskell.org/Functional_Reactive_Programming). Evan Czaplicki, the creator of [Elm](http://elm-lang.org), also gives a great overview of FRP in his talk [Controlling Time and Space: Understanding the Many Formulations of FRP](https://www.youtube.com/watch?v=Agu6jipKfYw).
```
sebadoom: I would not say Haskell *is* a FRP language, rather it is a language that can conveniently express FRP due to its functional, pure and lazy nature.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+We'll conclude with another excellent quote from the first edition of [Eloquent JavaScript](http://eloquentjavascript.net/1st_edition/):
+
+> "A student had been sitting motionless behind his computer for hours, frowning darkly. He was trying to write a beautiful solution to a difficult problem, but could not find the right approach. Fu-Tzu hit him on the back of his head and shouted '_Type something!_' The student started writing an ugly solution. After he had finished, he suddenly understood the beautiful solution."
+>
+> —_Marijn Haverbeke, [Eloquent JavaScript, 1st Edition, Chapter 6](http://eloquentjavascript.net/1st_edition/chapter6.html)_
+
+The concepts necessary for understanding <a href="#functional-programming" target="_self">functional programming</a>, <a href="#reactive-programming" target="_self">reactive programming</a>, and <a href="#functional-reactive-programming" target="_self">functional reactive programming</a> can be difficult to grasp, let alone _master_. Writing code that takes advantage of a paradigm's fundamentals is the initial step, even if it isn't entirely faithful at first. Practice illuminates the path ahead and also reveals potential revisions.
+
```
sebadoom: I like this 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+function impureHalf() {
+  return someNum / 2;
+}
+```
+
+In summary:
+
+* In pure functions, the same input will always produce the same output.
+* Pure functions rely only on local state and do not mutate external state.
+* Pure functions do not produce [side effects](https://en.wikipedia.org/wiki/Side_effect_%28computer_science%29).
```
kmaida: 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+
+Pure functions only depend on what's passed to them. For example, a pure function cannot reference variables from a parent scope unless they are explicitly passed into the function as arguments. Even then, the function can _not modify_ the parent scope.
+
+```js
+var someNum = 8;
+
+// this is NOT a pure function
+function impureHalf() {
+  return someNum / 2;
+}
```
kmaida: I think there might be something I'm missing / not understanding here. Is `pureFunction(val)` actually pure if it calls an impure function that was declared outside its scope?

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+
+* [Pure versus impure functions](https://toddmotto.com/pure-versus-impure-functions)
+* [Master the JavaScript Interview: What is a Pure Function?](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-a-pure-function-d1c076bec976#.kt48h2bfa)
+* [Functional Programming: Pure Functions](https://www.sitepoint.com/functional-programming-pure-functions/)
+
+---
+
+## <span id="state"></span>State
+
+**State** refers to the stored information a program has access to at a point in time. The contents of variables in an application at any given instant is the application's _state_.
```
kmaida: 👍 

I updated this section to:

```
**State** refers to the information a program has access to and can operate on at a point in time. This includes data stored in memory as well as OS memory, input/output ports, database, etc. For example, the contents of variables in an application at any given instant are representative of the application's _state_.
```

WDYT?

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+
+---
+
+## <span id="immutable-mutable"></span>Immutability and Mutability
+
+The concepts of **immutability and mutability** are slightly more nebulous in JavaScript than in some other programming languages. However, you will hear a lot about immutability when reading about <a href="#functional-programming" target="_self">functional programming</a> in JS. It's important to know what these terms mean classically and also how they are referenced and implemented in JavaScript. The definitions are simple enough:
+
+### Immutable
+
+If an object is **immutable**, its state cannot be modified after creation.
```
kmaida: 👍 Changed to value, since it's less dependent on another glossary term that way.

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+```js
+var str = 'Hello!';
+var anotherStr = str.substring(2);
+// result: str = 'Hello!' (unchanged)
+// result: anotherStr = 'llo!' (new string)
+```
+
+Using the `.substring()` method on our `Hello!` string does _not_ modify the original string. Instead, it creates a _new_ string. We could reassign the `str` _variable value_ to something else, but once we've created our `Hello!` string, it will always be `Hello!`.
+
+Numbers are immutable as well. The following will always have the same result:
```
kmaida: 👍 Agree!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+* [Immutable Objects with Object Freeze](http://adripofjavascript.com/blog/drips/immutable-objects-with-object-freeze.html)
+* [Mutable vs Immutable Objects](https://www.interviewcake.com/concept/java/mutable)
+* [Using Immutable Data Stuctures in JavaScript](http://jlongster.com/Using-Immutable-Data-Structures-in-JavaScript)
+* [Getting Started with Redux](https://egghead.io/courses/getting-started-with-redux) (includes examples for addressing immutable state)
+
+---
+
+## <span id="imperative-declarative"></span>Imperative and Declarative Programming
+
+While some languages were designed to be **imperative** (C++, PHP) or **declarative** (SQL, HTML), JavaScript (and others like [Java](http://openjdk.java.net/projects/lambda/) and [C#](https://msdn.microsoft.com/en-us/library/bb534803(v=vs.110).aspx)) can support both programming paradigms.
```
kmaida: 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
kmaida: As much as possible, I wanted to stick with JS for all examples because the article is specifically about how the terms apply to JS. While I totally agree that SQL is more ideal for a general definition of declarative, I was hoping to create two pieces of code doing the same thing in the same language to demonstrate imperative vs. declarative and the idea that they're both present in JS. I've seen a few examples using map to demo declarative in JS (like [this](http://latentflip.com/imperative-vs-declarative) and [this](http://codenugget.co/2015/03/05/declarative-vs-imperative-programming-web.html) so I'd like to stick with this example. I could maybe also add a SQL example too, but I'd rather not ditch the JS example. WDYT?

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
kmaida: 👍

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
sebadoom: No, I was restating it is in fact impure. Pure functions cannot make calls to impure functions. Point is: sometimes one might think something is pure and call it only to realize it was in fact impure. This is the problem with languages that cannot enforce purity through its type system, and JavaScript is one of them.

kmaida: 👍 Made some changes after our convo, thanks!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
```diff
@@ -0,0 +1,740 @@
...
+
+There are also libraries available to support immutability in JS. [Mori](http://swannodette.github.io/mori/) delivers persistent data structures based on Clojure. [Immutable.js](https://facebook.github.io/immutable-js/) by Facebook also provides [immutable collections for JS](https://auth0.com/blog/intro-to-immutable-js/). Utility libraries like [Underscore.js](http://underscorejs.org) and [lodash](http://www.lodash.com) provide methods and modules to promote a more immutable [functional programming _style_](https://github.com/lodash/lodash/wiki/FP-Guide).
+
+### Immutability and Mutability Takeaways
+
+Overall, JavaScript is a very mutable language. Some styles of JS coding _rely_ on this innate mutability. However, when writing functional JS, implementing immutability requires mindfulness. JS will not natively throw errors when you modify something unintentionally. Testing and libraries can assist, but working with immutability in JS takes practice and methodology.
+
```
kmaida: 👍 Added this and a link to persistence. Thanks!

_posts/2017-02-10-glossary-of-modern-javascript-concepts.markdown
kmaida: 👍 Added a SQL example (also kept JS example to stay on JS topic too).

PR #662, opened 2017-03-06
Managing State in Angular with ngrx/store

kmaida:
Fix weird merge issues. Re-establish branch. Resubmit PR.

### DIFF
--- a/_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
+++ b/_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
@@ -292,7 +292,7 @@ The [`ng g` command (or its longform, `ng generate`)](https://github.com/angular
 
 ## Create a Routing Module
 
-Let's build a separate `NgModule` to support routing. Create a new file in the `src/app/core` folder called `app-routing.module.ts`:
+Let's build a separate [`NgModule`](https://auth0.com/blog/angular-2-ngmodules/) to support routing. Create a new file in the `src/app/core` folder called `app-routing.module.ts`:
 
 ```typescript
 // src/app/core/routing-module.ts
@@ -1620,7 +1620,7 @@ Now our app feels more personalized.
 
 We can log in and out of our app, but that doesn't offer much more than simple personalization at the moment. Any visitor can still navigate to any route they wish if they simply enter URLs manually. Let's implement a route guard so that routes are activated only for logged in users.
 
-> **Important Security Note:** In our simple demo app, authentication is simply for routing because we don't have a server component. _Client-side authentication does not confer security features._ If you're building an authenticated app with a server, you'll need to authorize API requests with the JWT provided by Auth0 using a `Bearer` header. You can read more on how to do this in the [Auth0 Angular 2 Calling APIs docs](https://auth0.com/docs/quickstart/spa/angular2/08-calling-apis). The [`angular2-jwt` package](https://github.com/auth0/angular2-jwt) we installed provides `AUTH_PROVIDERS` to help accomplish this. When making API calls in an authenticated app, we would secure our server requests _in addition to_ implementing presentational route guards.
+> **Important Security Note:** In our simple demo app, authentication is simply for routing because we don't have a server component. _Client-side authentication does not confer security features._ If you're building an authenticated app with a server, you'll need to authorize API requests with the JWT provided by Auth0 using a `Bearer` header. You can read more on how to do this in the [Auth0 Angular 2 Calling APIs docs](https://auth0.com/docs/quickstart/spa/angular2/08-calling-apis). The [`angular2-jwt` package](https://github.com/auth0/angular2-jwt) we installed provides `AUTH_PROVIDERS` to help accomplish this. When making API calls in an authenticated app, we would secure our server requests _in addition to_ implementing presentational route guards. You can read more about securing a Node API in the [Angular 2 Authentication tutorial here](https://auth0.com/blog/angular-2-authentication/).
 
 Create a new file in `src/app/core` called `auth.guard.ts`:
 
--- a/_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
+++ b/_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
@@ -474,24 +474,24 @@ export function petTagReducer(state: PetTag = initialTag, action: Action) {
         complete: action.payload
       });
     case RESET:
-      return Object.assign({}, state, action.payload);
+      return Object.assign({}, state, initialTag);
     default:
       return state;
   }
 }
 ```
 
-We need to import `Action` from ngrx/store. Then we need the `PetTag` model and its default state, `initialTag`. We also need to import the actions we created in the previous step.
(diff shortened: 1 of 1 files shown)


_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+          }
+          localStorage.setItem('profile', JSON.stringify(profile));
+          this.userProfile = profile;
+
+          // on successful authentication and profile retrieval, go to /create route
+          this.router.navigate(['/create']);
+        });
+      } else if (authResult && !authResult.idToken) {
+        // authentication failed
+        throw Error(`There was an error authenticating: ${authResult}`);
```
sebadoom: Shouldn't this just show the login screen or remove the invalid credentials rather than throw an error?

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+</p>
+...
+{% endraw %}
+{% endhighlight %}
+
+Now our app feels more personalized.
+
+### Create a Route Guard
+
+We can log in and out of our app, but that doesn't offer much more than simple personalization because none of our routes are _protected_ yet. Unauthenticated users can still access everything if they enter URLs manually. Let's implement a route guard so that only logged in users can create and customize pet tags.
```
sebadoom: I'd note that this is just for routing purposes in this specific case. Any application requiring authentication must validate the token server side. In this case there is no server component, so authentication is just for routing.

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+
+Create a new file in `src/app/core` called `auth.guard.ts`:
+
+```typescript
+// src/app/core/auth.guard.ts
+import { Injectable } from '@angular/core';
+import { Router, CanActivate } from '@angular/router';
+import { AuthService } from './auth.service';
+
+@Injectable
```
sebadoom: I'd be cautious about the wording here. This gives the impression that whatever routes are protected by `AuthGuard` are not accessible. They are, it's just it is not convenient to do so without changing the code here. I'd remark that client side authentication is just for presentation purposes and does not confer any security or access control features.

sebadoom: I'd be careful about the wording here. I'd remark that client side authentication is just for presentation purposes and does not confer any security features (i.e. access is not restricted).

sebadoom: I'd be careful about the wording in this paragraph as to prevent a false sense of security. Client-side authentication is just for presentation purposes and does not confer any security features (in other words, routes protected with `AuthGuard` in this example are still accessible, though visual elements enabling so are hidden).

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+
+The `TOGGLE_CLIP` and `TOGGLE_GEMS` actions toggle booleans that are assigned in the `initialTag` state. Therefore, we don't need a payload when we dispatch these actions; we can simply set the value to its opposite in these cases.
+
+We're sending a payload with the `COMPLETE` action because we want to explicitly set it to `true`, and only do so once for each tag created. We could use a toggle for this as well, but for clarity, we'll dispatch a specific value as a payload instead.
+
+> **Note:** Notice that the `RESET` case still uses `action.payload`, although we have _access_ to the `initialTag` constant from imports. This is to maintain purity. We aren't passing `initialTag` into the reducer _except_ as the default value of the `state` argument. This means that if we were to return `Object.assign({}, state, initialTag)` for `RESET`, the function would no longer be pure: it would utilize data outside its own scope. Instead, we'll pass `initialTag` as the payload when dispatching the `RESET` action in our _component_ rather than in the reducer.
```
kmaida: 👍 Got it, thanks for the info! Updating the code and the side note to address this!

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+      return Object.assign({}, state, action.payload);
+    default:
+      return state;
+  }
+}
+```
+
+We need to import `Action` from ngrx/store. Then we need the `PetTag` model and its default state, `initialTag`. We also need to import the actions we created in the previous step.
+
+Now we'll create our `petTagReducer()` function. The reducer accepts previous `state` and the dispatched `action` as arguments. Remember that this is a _pure_ function: inputs determine outputs and the function does not modify global state. This means that when we return anything from the reducer, it either needs to be a new object or it can output an unmodified input (such as in the `default` case).
```
kmaida: 👍

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
```diff
@@ -0,0 +1,1722 @@
...
+
+We're going to use [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) to return new objects containing the values from source objects in most cases. The sources will be the _previous state_ and objects containing the action _payload_.
```
kmaida: 👍

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
kmaida: It's an error because something beyond invalid credentials went wrong here (Lock handles invalid credentials before this would be triggered). Something went wrong communicating with the Auth0 server, etc. I didn't want to go too deep into it but did want to output the resulting error so it could be debugged by the developer. (To be frank, I've never personally seen this happen.) I can update it to log the result instead of throwing an error, and show the Lock widget again though.

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
kmaida: 👍

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
kmaida: The three comments above all refer to the same thing, so 👍 to all :)

kmaida:
Pending SEO review.

_posts/2017-03-07-managing-state-in-angular-with-ngrx-store.markdown
sebadoom: Oh, OK, I was wondering what the behavior of this would be given the presence of Angular (does it catch the error somewhere? Or does it just fail and ends up with a broken page?). In any case, as you said, it won't matter as the developer will probably want to handle it in some application specific way 👍

PR #670, opened 2017-03-10
Access Requests Post

astanciu:
Do not publish yet.

Also, is it possible to set the header to a full image, like here:
https://medium.com/@astanciu/acces-requests-ecb4301e8032


diegopoza:
@astanciu nope, you cannot set the header to a full image, we can only set an icon that design creates for us

astanciu:
ok thanks. Is it possible to publish this in a "preview" mode, only visible via direct url?

PR #678, opened 2017-03-13
[MERGE ME ON 03/28] Ethereum, part 2 draft

sebadoom:
eth1
eth2
eth3


_posts/2017-03-14-an-introduction-to-ethereum-and-smart-contracts-part-2.markdown
```diff
@@ -0,0 +1,499 @@
...
+
+[Bitcoin](https://www.bitcoin.com) took the world by suprise in the year 2009 and popularized the idea of decentralized secure monetary transactions. The concepts behind it, however, can be extended to much more than just digital currencies. [Ethereum](https://www.ethereum.org) attempts to do that, marrying the power of decentralized transactions with a Turing-complete contract system. In this post we will take a closer look at how Ethereum works and what makes it different from Bitcoin and other blockchains. Rean on!
+
+This is post 2 from a three-post series about Ethereum. [Read post 1 if you haven't done so.](https://auth0.com/blog/an-introduction-to-ethereum-and-smart-contracts/)
+
+{% include tweet_quote.html quote_text="Ethereum marries the power of decentralized transactions with Turing-complete contracts!" %}
+
+-----
+
+## Introduction
```
brunokrebs: This is indeed an interesting topic, but in my opinion the article takes to long to show the real power of these smart contracts. Maybe you could add some tale, some example, of what is possible to do with this technology in the beginning of the article.

This would encourage the reader to keep reading the whole article (which is quite lengthy and not that easy to grasp).

By the way, I have added this comment here because I would prefer to see an example around here, but I already read the whole article.

PR #703, opened 2017-03-31
Build Better Universal JavaScript Apps with Next.js 2.0

unicodeveloper:


_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+
+We can easily set up authentication in a **Next.js 2.0** apps by using the [Lock Widget](https://auth0.com/lock). If you don't already have an Auth0 account, [sign up](javascript:signup\(\)) for one now. Navigate to the Auth0 [management dashboard](https://manage.auth0.com/), click on `New client` by the right hand side, select Regular Web App from the dialog box and then go ahead to the `Settings` tab where the client ID, client Secret and Domain can be retreived.
+
+**Note:** Make sure you set the  `Allowed Callback URLs` to `http://localhost:3000/` or whatever url/port you are running on. Also set the `Allowed Origins (CORS)` to `http://localhost:3000/` or whatever domain url you are using, especially if it is hosted.
+
+Authentication in a Next.js app could be a little complicated because you have to ensure that the server-rendered pages are authenticated, meaning they need to have access to the token.
+
+In the example below, the token returned from Auth0 is stored in localStorage and also as a cookie.
```
sebadoom: Change localStorage to Local Storage or LocalStorage and put a link to the MDN page about it.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+      params: {
+        scope: 'openid profile email',
+        state: secret
+      }
+    }
+  }
+}
+
+export const show = (container) => getLock(getOptions(container)).show()
+export const logout = () => getLock().logout({ returnTo: getBaseUrl() })
```
sebadoom: This appears to fail.
screenshot from 2017-04-04 20-18-02

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+const getLock = (options) => {
+  const config = require('../config.json')
+  const Auth0Lock = require('auth0-lock').default
+  return new Auth0Lock(config.AUTH0_CLIENT_ID, config.AUTH0_CLIENT_DOMAIN, options)
+}
+
+const getBaseUrl = () => `${window.location.protocol}//${window.location.host}`
+
+const getOptions = (container) => {
+  const secret = uuid.v4()
```
sebadoom: I'm a bit lost here, what is this secret used for? It is not the secret with which the JWT is signed, what is it used for?

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+
+_pages/auth/signed-in.js_
+
+```js
+
+import React, { PropTypes } from 'react'
+
+import { setToken, checkSecret, extractInfoFromHash } from '../../utils/auth'
+
+export default class SignedIn extends React.Component {
```
sebadoom: I'm missing where this class is used later on.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+export const getUserFromCookie = (req) => {
+  if (!req.headers.cookie) {
+    return undefined
+  }
+  const jwtCookie = req.headers.cookie.split(';').find(c => c.trim().startsWith('jwt='))
+  if (!jwtCookie) {
+    return undefined
+  }
+  const jwt = jwtCookie.split('=')[1]
+  return jwtDecode(jwt)
```
sebadoom: This **DOES NOT** validate the token's signature! In fact, I couldn't find the signing key anywhere in the project!

I just built a forged token and tested this. It takes any JWT. See my name in the page.

screenshot from 2017-04-04 20-37-36

This code is used here: https://github.com/auth0-blog/next2-auth0/blob/master/hocs/defaultPage.js#L9

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+  const {id_token, state} = getQueryParams()
+  return {token: id_token, secret: state}
+}
+
+export const setToken = (token) => {
+  if (!process.browser) {
+    return
+  }
+  window.localStorage.setItem('token', token)
+  window.localStorage.setItem('user', JSON.stringify(jwtDecode(token)))
```
sebadoom: Why are cookies and local storage used at the same time? Is it to make it easier to get the data from the cookie?

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+  }
+}
+
+```
+
+**Note:** The `next.config.js` file is a regular Node.js module.
+
+### 8. Composed CSS support
+
+Before now, `next/css` was the default `CSS-in-JS` solution for Next.js. In **Next.js 2**, it has been deprecated in favor of [styled-jsx](https://github.com/zeit/styled-jsx), a Babel transformation that provides full, scoped and component-friendly CSS support for JSX (rendered on the server or the client).
```
unicodeveloper: I couldn't find any links. It was just announced in their release statement.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+const app = next({ dev })
+const handle = app.getRequestHandler()
+
+app.prepare().then(() => {
+
+const server = express()
+
+  server.get('/p/:username', (req, res) => {
+    return app.render(req, res, '/z', {
+      ...req.query,
```
unicodeveloper: This is the rest operator. It is used to destructure `req.query`. `req.query` comes in form of an object.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
```diff
@@ -0,0 +1,556 @@
...
+* Serving of static files
+* CSS isolation and modularization
+
+So, what has changed? What improvements have been made? Are there new features? Will your Next.js apps built with this new release be truly production ready? Let's go through everything that **Next.js 2.0** has to offer.
+
+### 1. Faster Compilation Times
+
+Every developer and their dog need their apps to be highly performant from development stage up until production. A lot of work has been done to improve the dev build time that **Next.js 1.0** brought to the scene. With great joy, I hereby announce to you that **Next.js 2** comes bundled with shorter build/rebuild times. This was made possible by offering [Lazy compilation during development](https://github.com/zeit/next.js/pull/1111)
+
+![Lazy Compilation during development](https://cdn.auth0.com/blog/next20/lazycompilationdev.png)
```
unicodeveloper: Before now, When you run `next`, it compiles all the pages. But now, lazy compilation ensures that it is only when a user hits a page, that compilation happens. So each page that is called by the user is an on-demand entry

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: So @luisrudge wrote the original code for this. Cookies were used because it could be stored on the server. So if authentication is happening on a server-side rendered page, it uses the information from cookies else if it's happening client side, it grabs the user info from LocalStorage.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: This is very interesting. I need to check it out.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: So in Next.js, you can use a React Component or a simple function. This class was just given a name. I could also decide to make the class anonymous. Any of them works

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: I will check this out

luisrudge: The `logout` method is not documented and wasn't working properly after we upgraded auth0js to v8 (more info [here](https://github.com/auth0/lock/issues/968)). You have to either rollback to v10.7.2 or wait until the next release (10.15) where [this PR](https://github.com/auth0/lock/pull/966) will fix the issue.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: Adding a name to the class makes it easier to debug the component in react dev tools

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: This was just to add to the `state` option. It can be removed when you update the package because we now create a new random string when `state` is empty

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: I didn't validate anything. You have to create a regular webserver to do that, which I didn't. If you add the secret to your nextjs app, it will leak the secret to the front end.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: related: https://github.com/luisrudge/next.js-auth0/issues/10

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: @luisrudge The auth0 package?

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: So @luisrudge . I just discovered that you can have a `next.config.js` file and configure it to load environment files like so:

```js
const webpack = require('webpack');

require('dotenv').config();

module.exports = {
  webpack: (config) => {
    config.plugins.push(
      new webpack.DefinePlugin({
        'process.env.AUTH0_SECRET': JSON.stringify(process.env.AUTH0_SECRET)
      })
    );

    return config;
  },
};
```

So, it will load the value of `process.env.AUTH0_SECRET` on the server and on the client, once you have a `.env` file that loads the value of the secret. 

How about us taking that approach?

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: Alright, I have fixed this @sebadoom

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: @sebadoom How were you able to inject the forged token. I am trying to replicate your process here?

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: Yes. auth0-js has this by default: https://github.com/auth0/auth0.js/blob/b6f1c57098021a54e5a68ecc01c1fbaf88d4fa75/src/web-auth/transaction-manager.js#L35

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
luisrudge: @unicodeveloper if you do that, you'll leak the secret in the web app. the secret has to be available only in the server, so you have to use a custom express server to do that.

unicodeveloper: Sorry @luisrudge How would it still leak the secret when I am loading it from an environment variable? No one gets to see the secret. 

or do you mean someone can log `process.env.AUTH0_SECRET` in the browser and still get access to it?

luisrudge: if you load the secrets into the webapp (no matter if you use constants, env variables etc), it will be available in the web app. `process.env.AUTH0_SECRET` will be a public variable.

unicodeveloper: Hmmmn, thanks. Will have to research how to use a custom express server to handle that

luisrudge: Here's an example of how I use the webpack define plugin in  a personal project:

```js
new webpack.DefinePlugin({
  'process.env': {
    'NODE_ENV': JSON.stringify(isProd ? 'production' : 'development')
  },
  '__API_URL__': JSON.stringify('http://myapi.com'),
  '__AUTH0_DOMAIN__': JSON.stringify(process.env.AUTH0_DOMAIN),
  '__AUTH0_CLIENTID__': JSON.stringify(process.env.AUTH0_CLIENTID),
})
```

All those variables are available in my web app. If you want to use the secret, you have to follow this example: https://github.com/zeit/next.js/tree/master/examples/custom-server-express

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
sebadoom: Exactly, this is a typical question of someone not familiar with Next. A short explanation like yours right here would be appropriate for the article.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
sebadoom: No, this is the [spread operator](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator) ([rest](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/rest_parameters) is for function arguments). Point is it is not clear how the spread operator destructures or expands what is contained in req.query. In particular req.query is a dictionary, so the use of spread syntax here is unclear. I believe this works because Next uses babel internally, the use of spread syntax for objects is not valid ES6. In fact, it is a stage-3 proposal. https://github.com/tc39/proposal-object-rest-spread

unicodeveloper: Yeah, Next uses Babel internally.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
unicodeveloper: I'll provide that.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
sebadoom: I am not familiar with Next, but this is exactly my point, though. Next blurs the line between what is server side and what is client side. This means that extra care should be taken when handling secrets. Authentication requires handling secrets. So if we are going to show how to do useful auth here, it should be clear how to handle secrets in Next (failing to do this could result in people inadvertently publishing their secrets to clients!).

In other words, sure, if you are just doing stuff client-side, there is little point in checking the signature of the id_token (although per OIDC you should, which is why RS256 would be a good option here), but it is disingenuous to show how to do auth for a Next app and at no point show how things could or should be validated.

_posts/2017-03-29-build-better-universal-apps-with-nextjs2.markdown
sebadoom: I meant I could not find any uses for this component in the project. The name or class use is not important. Is it used anywhere?

PR #712, opened 2017-04-04
Personalize user experience

brunokrebs:


brunokrebs:
It is missing icon and bg color

PR #738, opened 2017-04-11
Add a security note stating where to look for instructions on securin…

kmaida:
…g an API with Auth0.

The Polymer post uses the Chuck Norris homegrown API and homegrown authentication for its demo, which does not use Auth0 for security. Auth0 is only used in the aside to demonstrate client-side protection. A note has been added with instructions on how to implement APIs and implicit grant, but the demo does not need to be updated as it does not use Auth0 to call an API.


kukicado:
Reviewed. Looks good.

PR #739, opened 2017-04-11
Add security note about APIs to Elm post.

kmaida:
The Elm post uses the Chuck Norris API and homegrown authentication in its demo. The Auth0 aside shows how to integrate Lock on the client-side only. An important security note has been added with links informing the reader how to set up APIs if needed and use implicit grant.


kukicado:
Reviewed. Looks good.

PR #744, opened 2017-04-11
Update Angular 2 Auth Post to use access_token

kukicado:


_posts/2016-09-29-angular-2-authentication.markdown
```diff
@@ -707,13 +819,81 @@ We need to update the call to the `/api/deals/private` to include our JWT. There
...
+**Note:** The **authHTTP** library currently defaults to sending the `id_token` to the backend. Our API requires the `access_token` instead, but we can easily change this. The `authHTTP` library can be customized and we'll do this by creating a custom factory and changing the default token type from an `id_token` to an `access_token`. To do this, create a new file and call it `auth.factory.ts`. Our implementation will be as follows:
+
+```js
+import { Http, RequestOptions } from '@angular/http';
+import { AuthHttp, AuthConfig } from 'angular2-jwt';
+
+export function authHttpServiceFactory(http: Http, options: RequestOptions) {
+  return new AuthHttp(new AuthConfig({
+    tokenName: 'token',
+    tokenGetter: (() => localStorage.getItem('access_token'))
```
brunokrebs: If I'm not mistaken, version `0.2.2` of `angular2-jwt`, released yesterday, uses `token` now as the default token name and to retrieve the token to be appended in the `Authorization` header. So I think that using `token` instead of `access_token` will be easier and make this factory useless...

[V 0.2.2](https://github.com/auth0/angular2-jwt/blob/0.2.2/angular2-jwt.ts)

brunokrebs: Oh, and looking a little bit further, version 1.X of this library used to have `DEFAULT_TOKEN_NAME = 'id_token';`, then on the first minor releases of version 2.X, they changed it to `DEFAULT_TOKEN_NAME = 'access_token';` and now, 22 hours ago, they've changed it to `DEFAULT_TOKEN_NAME = 'token';`

kukicado: Interesting. I think for this article I'll leave it as is (since all the other dependencies are older) and in the ANgular 4 post I'll use the latest version of all the libs. Make sense @brunokrebs ?

brunokrebs: Hmm, you are instructing the reader to issue `npm install angular2-jwt --save` which will give them the last version, and therefore it won't make sense to write this factory. 

Furthermore, you state that `The authHTTP library currently defaults to sending the id_token to the backend`, which I think it is not write. The current library sends `token` not `id_token`.

kukicado: Lol i think they changed it right after I wrote the article. Let me see what happens when I remove my factory (since it may be redundant now).

kukicado:
@brunokrebs good call - after updating the library to latest I was able to make it work without having to write the custom factory. Thanks!

Post should be ready to merge now.

_posts/2016-09-29-angular-2-authentication.markdown
```diff
@@ -819,67 +819,7 @@ We need to update the call to the `/api/deals/private` to include our `access_to
   }
 ```
 
-**Note:** The **authHTTP** library currently defaults to sending the `id_token` to the backend. Our API requires the `access_token` instead, but we can easily change this. The `authHTTP` library can be customized and we'll do this by creating a custom factory and changing the default token type from an `id_token` to an `access_token`. To do this, create a new file and call it `auth.factory.ts`. Our implementation will be as follows:
```
brunokrebs: It worked?! 💪

kukicado: Yup :)

_posts/2016-09-29-angular-2-authentication.markdown
```diff
@@ -677,15 +706,98 @@ export class AppComponent {
   title = 'Daily Deals';
 
   // We'll need to include a reference to our authService in the constructor to gain access to the API's in the view
-  constructor(private authService: AuthService) {}
+  constructor(private authService: AuthService) {
+    // Generate a nonce
+    this.nonce = this.authService.generateNonce();
```
brunokrebs: Does this work? I mean, there is no definition of `this.nonce`.

PR #750, opened 2017-04-14
Replace outdated aside with new flow.

kmaida:
Replaced the old (outdated) Angular aside with the new generic one: https://github.com/auth0-blog/angular-auth0-aside/blob/master/angular-auth0-aside.markdown


kmaida:
Aside has been reviewed by @kukicado

PR #755, opened 2017-04-17
Update Migrating an AngularJS App to Angular (part 3) Auth0 aside wit…

kmaida:
…h pipeline 2 information (added an important note and link to sample repo).

### DIFF
--- a/_posts/2016-11-14-migrating-an-angular-1-app-to-angular-2-part-3.markdown
+++ b/_posts/2016-11-14-migrating-an-angular-1-app-to-angular-2-part-3.markdown
@@ -663,6 +663,8 @@ declare var window: any;
 
 @Injectable()
 export class AuthService {
+  // IMPORTANT: id_token should only be used on the client, NOT to secure APIs.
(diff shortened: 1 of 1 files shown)


_posts/2016-11-14-migrating-an-angular-1-app-to-angular-2-part-3.markdown
```diff
@@ -722,17 +724,18 @@ export class AuthService {
...
-    // This searches for an item in localStorage with key == 'id_token'
-    return tokenNotExpired();
+    return tokenNotExpired('id_token');
   }
 
 }
 ```
 
-As always, first we'll talk about the imports. `Injectable` is necessary for any injectable service and is provided by the boilerplate. We'll want to use `Router` to redirect the authenticated user back to the route they logged in from. We also need `tokenNotExpired` from `angular2-jwt` to get a user's authentication state.
+As always, first we'll talk about the imports. `Injectable` is necessary for any injectable service and is provided by the boilerplate. We'll want to use `Router` to redirect the authenticated user back to the route they logged in from. We also need `tokenNotExpired('id_token')` from `angular2-jwt` to get a user's authentication state.
```
brunokrebs: Good to know that it has changed the default value.

PR #768, opened 2017-04-19
Add notes on using access token for your API - Pipeline 2

unicodeveloper:


_posts/2017-03-16-web-components-how-to-craft-your-own-custom-components.markdown
```diff
@@ -531,6 +531,9 @@ lock.on("authenticated", function(authResult) {
 
 ```
 
+**Note:** If you have an API for your application, the API should _always_ be secured. The [`id_token` should not be used to secure an API](https://auth0.com/blog/why-should-use-accesstokens-to-secure-an-api); instead use an `access_token` with the appropriate configuration. You can read about how to [implement API authentication with Auth0](https://auth0.com/docs/apis) with [implicit grant](https://auth0.com/docs/api-auth/tutorials/implicit-grant).
```
brunokrebs: Not sure if we could state that an API should always be secured. Furthermore, this note about the `id_token` got pretty popular in our updates, so I think we could stick to a default one.

The one that I liked the most is [this one from Kim](https://github.com/auth0/blog/pull/771/files). I like because it  links to the `Which OAuth Flow to Use`. But, it is up to you if you want to change this note or not.

unicodeveloper: I can change it to Kim's own ASAP

unicodeveloper:
@brunokrebs Done

PR #779, opened 2017-04-21
Add Important Security Note about APIs and Access Tokens

unicodeveloper:


_posts/2017-02-09-migrating-a-php5-app-to-7-part-three.markdown
```diff
@@ -424,7 +424,7 @@ $userInfo = $auth0Oauth->getUser();
 
 Then, we moved on to instantiating the `Authentication` class.
 
-The `$auth0->get_oauth_client()` method by default stores user information in the PHP session, and we also instructed it to save the `access_token` and `id_token`. 
+The `$auth0->get_oauth_client()` method by default stores user information in the PHP session, and we also instructed it to save the `refresh_token` and `id_token`.
```
brunokrebs: I don't get it. You will keep saving `id_token` in the backend even though we are preaching against this practice?

unicodeveloper: We are not preaching against saving `id_token` in the database. As long as you are using it for basic user info and UI rendering, it is fine. We are against using `id_token` for making API calls. Do you get the drift?

So in this app, I just used `id_token` to log the user in and display basic info. Nothin more.

brunokrebs: Hmm, I get it, you are using it just to build the interface in the server side. Ok, makes sense.

PR #786, opened 2017-04-25
Add Important Security Note about APIs and Access Tokens

unicodeveloper:


_posts/2015-12-28-securing-your-nodebots-with-json-web-tokens.markdown
```diff
@@ -163,7 +163,9 @@ var jwtCheck = jwt({
 });
 ```
 
-This way, you can easily allow users to log into your NodeBots interface, with jsut a few lines of javascript!
+**Important API Security Note:** If you want to use Auth0 authentication to authorize _API requests_, note that you'll need to use [a different flow depending on your use case](https://auth0.com/docs/api-auth/which-oauth-flow-to-use). Auth0 `idToken` should only be used on the client-side. [Access tokens should be used to authorize APIs](https://auth0.com/blog/why-should-use-accesstokens-to-secure-an-api/). You can read more about [making API calls with Auth0 here](https://auth0.com/docs/apis).
+
+This way, you can easily allow users to log into your NodeBots interface, with just a few lines of javascript!
```
kmaida: On the last line:

`javascript` -> "JavaScript"

unicodeveloper: Thanks

PR #801, opened 2017-04-27
Update to use access tokens and add important security note

unicodeveloper:


kmaida:
Reviewed, no comments. 👍

PR #832, opened 2017-05-16
Replacing images and adding some text on who uses Extend

glennblock:
introducing auth0 extend the new way to extend your saas


glennblock:
Ignore the button that looks out of what. It should be fine when merged.

@woloski @tjanczuk updated images.

PR #836, opened 2017-05-18
Implement Experimental Disclaimer Feature

kukicado:
Adds a flag called "experimental" that when set to "true" will display a disclaimer at the top of article saying that the post contains experimental code and should not be used in production.


diegopoza:
@kukicado looks good to me, please merge it

PR #848, opened 2017-05-31
ASP.NET Core auth

brunokrebs:


_posts/2017-05-23-asp-dot-net-core-authentication-tutorial.markdown
```diff
@@ -0,0 +1,733 @@
...
+  }
+}
+```
+
+This class, for the time being, contains only one public method, `Register`, which accepts HTTP POST requests with `Credentials` in the body. Besides this public method, it contains the following private methods:
+
+* `ConvertToUnixTimestamp`—which is used to fill `iat` (issued at), `nbf` (not before), and `exp` (expiration) claims in the JWTs generated. These claims are used to validate if the token is indeed valid and are expected to contain a time represented in seconds passed after 1970.
+* `Errors` and `Error`—which are used to send error messages as JSON.
+* `GetToken`—a method that receives a set of key values pairs and generates a JWT adding all the values received alongside with the `iss`, `aud`, `nbf`, `iat`, and `exp` claims.
+* `GetIdToken` and `GetAccessToken`—which are the methods that generate the `access_token` and the `id_token` for users. If you are wondering [what is the difference between these two tokens and when to use one or another, take a look here](2017-04-11-why-should-use-accesstokens-to-secure-an-api).
```
sebadoom: I was not aware putting partial URLs worked for our articles, cool!

sebadoom: I would also clarify that usually access tokens and id tokens have different audience claims (in contrast to what the code above does).

_posts/2017-05-23-asp-dot-net-core-authentication-tutorial.markdown
```diff
@@ -0,0 +1,733 @@
...
+    "password": "123#Bruno"
+}'  http://localhost:5000/api/account | jq -r '.access_token')"
+
+# Just showing the ACCESS_TOKEN contents
+echo $ACCESS_TOKEN
+
+# Passing the ACCESS_TOKEN as a Authorization header and quering the list of groceries
+curl -H "Authorization: Bearer $ACCESS_TOKEN" http://localhost:5000/api/grocerylist
+```
+
```
sebadoom: Note that due to the example using the same audience for both the id_token and the access_token, if the user were to send the id_token in this case it would still work (which might be confusing for some users).

_posts/2017-05-23-asp-dot-net-core-authentication-tutorial.markdown
```diff
@@ -0,0 +1,733 @@
+---
+layout: post
+title: "ASP.NET Core Authentication Tutorial"
+description: "Learn how to handle authentication on ASP.NET Core applications"
+date: 2017-05-23 16:58
```
brunokrebs: It is not going to be released today, but thanks.

_posts/2017-05-23-asp-dot-net-core-authentication-tutorial.markdown
brunokrebs: Going to change the audience.

brunokrebs: And no, the url does not work like that, @sebadoom. This is a typo. It will if I add `/blog/` though

PR #864, opened 2017-06-08
Kotlin RESTful API

brunokrebs:


_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+
+For starters, if you haven't done so yet, this is a good time to sign up for a [free Auth0 account](javascript:signup\(\)). Having an Auth0 account, the first thing that we must do is to [create a new API on the dashboard](https://manage.auth0.com/#/apis). An API is an entity that represents an external resource, capable of accepting and responding to protected resource requests made by clients. And this is exactly what the Kotlin app that we just built is, an API.
+
+![Creating an API on Auth0's dashboard](https://cdn.auth0.com/blog/kotlin-jwts/create-auth0-api.png)
+
+**Note**, if you don't have the APIs menu item, you can enable it by going to your [Account Settings](https://manage.auth0.com/#/account/advanced). In this page, click in the **Advanced** tab, scroll down until you see **Enable APIs Section** and flip the switch.
```
brunokrebs: I have to update this.

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+</dependency>
+
+<dependency>
+  <groupId>com.auth0</groupId>
+  <artifactId>auth0-spring-security-api</artifactId>
+  <version>1.0.0-rc.2</version>
+</dependency>
+```
+
+After that we could also remove the two properties that we've added to the `application.properties` file, as they won't be used anymore. Then, to issue and validate JWTs, we would need to add the following Maven dependency:
```
kmaida: `After that` needs a comma after it (introductory clause).

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+		@Id @GeneratedValue(strategy = GenerationType.AUTO)
+		var id: Long = 0,
+
+		var username: String = "",
+
+		var password: String = ""
+)
+```
+
+Nothing new over here. Just another data class that holds users' properties. After that we will create the `ApplicationUserRepository.kt` class in the `persistence` package, with the following code:
```
kmaida: Comma after `After that`

No comma after `package,`

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+
+import com.auth0.samples.kotlinspringboot.model.ApplicationUser
+import org.springframework.data.repository.CrudRepository
+
+interface ApplicationUserRepository : CrudRepository<ApplicationUser, Long> {
+	fun findByUsername(username: String): ApplicationUser?
+}
+```
+
+In this case, the only difference when compared to `CustomerRepository` is that we defined a method called `findByUsername`. This method will be used by our home made solution to find users by their usernames. Now the last class, `SignUpController.kt`, will be created in the `controller` package with the following code:
```
kmaida: It's more typical to use the term "homegrown" or "proprietary" to refer to custom software solutions. I think "home made" is perfectly understandable, but for future reference, or if you'd like to change it up sometimes, feel free to update some of these. :)

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+		applicationUserRepository.save(applicationUser)
+	}
+}
+```
+
+The only endpoint defined in this controller is the `signUp` one, which allows new users to register in our application. The sign in process and token validation will be handled in a different area, as we will see soon. Note that to keep our users' password secure even in an eventual data breach, we have used the `BCryptPasswordEncoder` class, that comes along with Spring Security, to encode all passwords.
+
+### Issuing and Validating JWTs with Kotlin
+
+Now that we have our `User` data class mapped and an endpoint to allow new users to register themselves, we need to allow these users to sign in and also validate JWTs before enabling them to interact with our API. To achieve this, we will create two filters and one class: `JWTAuthenticationFilter`, `JWTAuthorizationFilter`, and `UserDetailsServiceImpl`. The first one, which will be responsible for the sign in feature, will be created in a new directory file called `JWTAuthenticationFilter.kt` in the same package as the `WebSecurity` class. This file will have the following source code:
```
kmaida: `The first one,` -> "The first filter"

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+				.signWith(SignatureAlgorithm.HS512, SECRET)
+				.compact()
+		res.addHeader(HEADER_STRING, TOKEN_PREFIX + " " + JWT)
+	}
+}
+```
+
+This filter defines two functions:
+
+- `attemptAuthentication`, which will parse the credentials from the user and try to authenticate it
```
kmaida: `authenticate it` -> makes it sound like the user is being called "it"; this should probably be "authenticate them"

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+- and `successfulAuthentication`, which will generate a JWT to users if they successfully authenticated.
```
kmaida: `JWT to users if they successfully authenticated` -> "JWT if the user is successfully authenticated"

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+
+Note that this filter, as the other one that we will create, uses some undefined constants like `SECRET` and `EXPIRATION_TIME`. To define these constants we will create a file called `SecurityConstants.kt` in this same directory with the following code:
```
kmaida: `Note that this filter, as the other one that we will create, uses` -> "Note that both filters use"

`To define these constants` needs a comma after it

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+				UsernamePasswordAuthenticationToken(user, null, emptyList<GrantedAuthority>())
+			else
+				null
+		}
+		return null
+	}
+}
+```
+
+This filter will get in action whenever a secured endpoint is requested and will start by checking if there is a token in the `Authorization` header. If it manages to find one, it will try to validate it and set the user in the `SecurityContext`. If no token is found, it will simply let the request move along the Spring Security filter chain, and then this request will be answered with a 401 (Unauthorized) status code.
```
kmaida: `This filter will get in action whenever` -> "This filter will be utilized whenever"

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+
+The last class that we will need to create is `UserDetailsServiceImpl`. This class will extend `UserDetailsService` class from Spring Security and will be responsible for finding users in our database so Spring Security can check their credentials. This class will be create in the main `kotlinspringboot` directory and will contain the following source code:
```
kmaida: `will extend` -> "will extend the"

`This class will be create in` -> "This class will be created in"

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+				.addFilter(JWTAuthorizationFilter(authenticationManager()))
+	}
+
+	override fun configure(auth: AuthenticationManagerBuilder?) {
+		auth!!.userDetailsService(userDetailsService).passwordEncoder(bCryptPasswordEncoder())
+	}
+}
+```
+
+After applying these changes, we can interact with our API again and check if its properly generating and validating JWTs:
```
kmaida: `check if its properly` -> "check if it's properly" (it's, conjunction of "it is")

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+curl -i -H "Content-Type: application/json" -X POST -d '{
+    "username": "admin",
+    "password": "password"
+}' http://localhost:8080/login
+
+# get customers passing the JWT contained by the Authorization header
+curl -H "Authorization: Bearer xxx.yyy.zzz" http://localhost:8080/customers
+```
+
+As you can see, to create our own security solution with JWTs is not that hard. But it requires much more work than what we had to do to integrate with Auth0. And keep in mind that we didn't even addressed more advanced topics like [Multifactor Authentication](https://auth0.com/docs/multifactor-authentication), [social identity providers](https://auth0.com/docs/identityproviders), [enterprise connections (Active Directory, LDAP, SAML, etc.)](https://auth0.com/enterprise), and etc. Having to handle such cases, would require us a lot more work. And even if we managed to deliver these features in a satisfactory speed, we wouldn't have as many security measures as we would while using Auth0.
```
kmaida: `to create our own security solution` -> "creating our own security solution"

`But it requires` -> "However, it requires"

`didn't even addressed more` -> "didn't even address more"

`, and etc.` -> ", etc." (don't use "and" with "etc.")

`Having to handle such cases, would require us a lot more work.` -> "Handling such cases would require a lot more work."

`features in a satisfactory speed` -> "features swiftly"

Note: try not to start sentences with "And" whenever possible. This is done twice in the above paragraph. Often we can just remove the "And" and start the sentence without it.

_posts/2017-06-07-developing-restful-apis-with-kotlin.markdown
```diff
@@ -0,0 +1,648 @@
...
+
+## Conclusion
+
+Using Kotlin in place of Java to write backend applications is easy, there are not that many pitfalls to be aware of. Kotlin integration with existing Java libraries is very good, as we could while using Spring Boot to write a RESTful API with Kotlin in this article, and the code developed was very concise and easy to read. We haven't covered that many features of Kotlin, but we managed to validate its applicability on the backend.
```
kmaida: `Using Kotlin in place of Java to write backend applications is easy, there are not that many pitfalls to be aware of. ` 

This contradicts what was said earlier in the article. You wrote: `Although this might be true, becoming a Kotlin developer and writing idiomatic code in this new language is not that straight forward.` Saying that using Kotlin instead of Java is easy now conflicts with the statement that it's _not_ straightforward. It might be worthwhile to consider removing one of these statements, or altering the conclusion a little bit so it's not contradictory.

`as we could while using Spring Boot to write a RESTful API with Kotlin in this article,` I'm not entirely sure what this means. Perhaps rewording this sentence would help.

brunokrebs: Not sure if I agree with your statement. Perhaps I used the wrong term but, as far as I know, `writing idiomatic` means writing in a way that is very specific to a language. That's why I started the sentence with `Although this might be true`.

What I mean is, writing in code in Kotlin is not that hard for a Java developer. Although becoming a true Kotlin developer, using its full power, its best features, in a very idiomatic(??) way, is not that straightforward.

Not sure if I managed to make myself clear with this explanation but, if so, could you please explain if I used idiomatic in a wrong way? Or perhaps how could I improve one of these sentences to better express what I explained here?

@kmaida

kmaida: Your use of `idiomatic` is correct. That's not the issue I see here. As the author, it might be transparent what you're saying, but as a _reader_, these statements seem to conflict and are confusing. I think it's important that you explain (in the conclusion) what you just said to explain this here in GitHub:

`writing in code in Kotlin is not that hard for a Java developer. Although becoming a true Kotlin developer, using its full power, its best features, in a very idiomatic(??) way, is not that straightforward.` (Though reword this to reduce commas and clauses for easier reading.)

This is in the introduction but _not_ in the conclusion, which makes the conclusion conflict with the introduction. I'd suggest just putting this in the conclusion too.

PR #889, opened 2017-06-23
Real-World Angular Series - Part 3

kmaida:
Publish date: July 5

### DIFF
--- a/_posts/2017-07-04-real-world-angular-series-part-3.markdown
+++ b/_posts/2017-07-04-real-world-angular-series-part-3.markdown
@@ -3,7 +3,7 @@ layout: post
 title: "Real-World Angular Series - Part 3: Fetching and Displaying API Data"
 description: "Build and deploy a real-world app with MongoDB, Express, Angular, and Node (MEAN): fetching, displaying, and filtering data."
 date: 2017-07-03 8:30
-category: Technical guide
+category: Technical guide, Angular, Angular 4
 banner:
   text: "Auth0 makes it easy to add authentication to your Angular application."
 author:
--- a/_posts/2017-07-04-real-world-angular-series-part-3.markdown
+++ b/_posts/2017-07-04-real-world-angular-series-part-3.markdown
@@ -23,6 +23,7 @@ tags:
 - node
 - mongodb
 - express
+- mean
 related:
 - 2017-06-27-real-world-angular-series-part-1
 - 2017-06-29-real-world-angular-series-part-2
--- a/_posts/2017-07-04-real-world-angular-series-part-3.markdown
+++ b/_posts/2017-07-04-real-world-angular-series-part-3.markdown
@@ -1,0 +1,823 @@
+---
+layout: post
+title: "Real-World Angular Series - Part 3: Fetching and Displaying API Data"
+description: "Build and deploy a real-world app with MongoDB, Express, Angular, and Node (MEAN): fetching, displaying, and filtering data."
+date: 2017-07-04 8:30
+category: Technical guide
+banner:
+  text: "Auth0 makes it easy to add authentication to your Angular application."
+author:
+  name: "Kim Maida"
+  url: "https://twitter.com/KimMaida"
+  mail: "kim.maida@auth0.com"
+  avatar: "https://en.gravatar.com/userimage/20807150/4c9e5bd34750ec1dcedd71cb40b4a9ba.png"
+design:
+  image: https://cdn.auth0.com/blog/angular/logo.png
+  image_size: "75%"
+  image_bg_color: "rgb(1, 70, 166)"
+  bg_color: "rgb(1, 70, 166)"
+  bg_merge: true
+tags:
+- javascript
+- angular
+- node
+- mongodb
+- express
+related:
+- 2017-06-27-real-world-angular-series-part-1
+- 2017-06-29-real-world-angular-series-part-2
+- 2016-11-07-migrating-an-angular-1-app-to-angular-2-part-1
+
+---
+
+**TL;DR:** This tutorial series covers building and deploying a full-stack JavaScript application from the ground up with hosted [MongoDB](https://www.mongodb.com/), [Express](https://expressjs.com/), [Angular (v2+)](https://angular.io), and [Node.js](https://nodejs.org) (also known as MEAN stack). The completed code is available in the [mean-rsvp-auth0 GitHub repo](https://github.com/auth0-blog/mean-rsvp-auth0/). **Part 3 of the tutorial series covers fetching data from MongoDB with a Node API and displaying and filtering it with Angular.**
+
+---
+
+## Part 3: Fetching and Displaying API Data
+
+The [second part of this tutorial](https://auth0.com/blog/real-world-angular-series-part-2) covered authentication, authorization, feature planning, and data modeling.
+
+The third installment in the series covers fetching data from MongoDB with a Node API and displaying and filtering the data with Angular.
+
+1. <a href="#api-events" target="_self">API: Fetching Events</a>
+2. <a href="#angular-api" target="_self">Angular: Fetching Events</a>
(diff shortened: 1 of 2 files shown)


_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+  ...
+```
+
+The code for this endpoint is very similar to the route fetching public events, but we'll include both the `jwtCheck` and `adminCheck` middleware. We'll also send `{}` as the query because we want to retrieve _all_ events in the database. We'll pass the same `_eventListProjection` to leave out locations and descriptions.
+
+> **Note:** It's worthwhile to note that this endpoint is simply for admin display purposes. We want the admin to be able to see and interact with a _listing_ of public and private events. However, authenticated _users_ can still see private event details too, they just need to know the direct link and can't access them from a list.
+
+### GET Event Details
+
+Now we'll fetch an event by ID with an `/api/event/:id` endpoint:
```
kukicado: with an => with the  (since the endpoint is singular right?)

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+        return res.status(400).send({message: 'Event not found.'});
+      }
+      res.send(event);
+    });
+  });
+  
+  ...
+```
+
+This authorized endpoint needs to have a parameter passed to it when called. The parameter should be the event ID so we can use the `findById()` method. If no event is found matching the ID we passed, we'll send a bad request error. Otherwise, we'll return the event.
```
kukicado: Instead of a bad request - should we send a 404 not found?

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+
+This is done differently for various value types. The search handles strings, JavaScript Date objects, and UTC strings. If we want to ensure that the search doesn't query certain properties, we'll make sure to pass them in as `excludedProps` when calling the method in our components.
+
+> **Note:** We won't search properties with values that are any other types because our RSVP app doesn't need this. If you'd like to see a more robust implementation that handles strings, numbers,  booleans, and dates, please check out [this filter-sort service Gist on GitHub](https://gist.github.com/kmaida/49dadc4c6c44116727ee859e21a32a46#file-filter-sort-service-ts).
```
kukicado: ,  booleans (extra space)

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+```js
+// server/api.js
+...
+/*
+ |--------------------------------------
+ | API Routes
+ |--------------------------------------
+ */
+ 
+ const _eventListProjection = 'title startDatetime endDatetime viewPublic';
```
kmaida: It's to indicate an internal variable in this case; only using `const` in the sense of ES6 because it doesn't change. Could also potentially be `const EVENT_LIST_PROJECTION` I suppose... do you feel it's very important?

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+- express
+- mean
+related:
+- 2017-06-27-real-world-angular-series-part-1
+- 2017-06-29-real-world-angular-series-part-2
+- 2016-11-07-migrating-an-angular-1-app-to-angular-2-part-1
+
+---
+
+**TL;DR:** This tutorial series covers building and deploying a full-stack JavaScript application from the ground up with hosted [MongoDB](https://www.mongodb.com/), [Express](https://expressjs.com/), [Angular (v2+)](https://angular.io), and [Node.js](https://nodejs.org) (also known as MEAN stack). The completed code is available in the [mean-rsvp-auth0 GitHub repo](https://github.com/auth0-blog/mean-rsvp-auth0/). **Part 3 of the tutorial series covers fetching data from MongoDB with a Node API and displaying and filtering it with Angular.**
```
kmaida: TBH I'm not sure I want to change this, since it's just explaining an acronym and I feel like it reads more smoothly without the word "the". Do you feel strongly about this?

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+* List of public events starting in the future
+* List of all public and private events (admin access required)
+* Event details (authentication required)
+* List of RSVPs associated with an event (authentication required)
+
+Open up the server `api.js` file and let's begin.
+
+### GET Future Public Events
+
+We'll start with an `/api/events` endpoint that retrieves all public events with a start date in the future from the `events` collection.
```
kmaida: The word "that" following the word "endpoint" then makes the suggested update sound strange...

_posts/2017-07-04-real-world-angular-series-part-3.markdown
kmaida: I used "an" because it's the first time a new endpoint has been introduced. "The" would be more appropriate if the user already had been introduced to this endpoint previously.

_posts/2017-07-04-real-world-angular-series-part-3.markdown
kmaida: Hmmm, I feel pretty strongly that it is a bad request. It means the request was somehow malformed, because it shouldn't actually be _possible_ to request an event that no longer exists if the app is performing the way it should be.

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+
+Before we start building out our components, let's make a utility service that we can build on throughout development.
+
+Run the following command to generate the boilerplate:
+
+```bash
+$ ng g service core/utils
+```
+
+We'll begin using this service to add an `isLoaded()` utility. Then we'll create methods to manage the display of event dates. Each event has a start datetime and an end datetime. Start and end  dates for a single event may be different days _or_ the same day. We want a way to collapse same-day events into one date when displaying them in the UI. We also don't need to show times on the main listings, only on detail pages. Finally, we'll want a way to determine if an event already happened and is now in the past.
```
kmaida: 👍

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+    ...,
+    DatePipe,
+    UtilsService
+  ],
+  ...
+})
+...
+```
+
+We're now ready to use our new utilities in components. We'll add more methods to this handy service as we need them throughout development.
```
kmaida: I don't really agree with a double use of "our" in a single sentence here.

_posts/2017-07-04-real-world-angular-series-part-3.markdown
kmaida: These double spaces are actually really weird because they don't show up in my markdown editor, haha. They happen on linebreaks and if I delete a space, it becomes a single run-on word. Very weird.

_posts/2017-07-04-real-world-angular-series-part-3.markdown
```diff
@@ -0,0 +1,824 @@
...
+        }
+        res.send(eventsArr);
+      }
+    );
+  });
+  
+  ...
+```
+
+The code for this endpoint is very similar to the route fetching public events, but we'll include both the `jwtCheck` and `adminCheck` middleware. We'll also send `{}` as the query because we want to retrieve _all_ events in the database. We'll pass the same `_eventListProjection` to leave out locations and descriptions.
```
kmaida: 👍

PR #892, opened 2017-06-23
Real-World Angular Series - Part 6

kmaida:
Publish: July 13

### DIFF
--- a/_posts/2017-07-13-real-world-angular-series-part-6.markdown
+++ b/_posts/2017-07-13-real-world-angular-series-part-6.markdown
@@ -3,7 +3,7 @@ layout: post
 title: "Real-World Angular Series - Part 6: Reactive Forms and Custom Validation"
 description: "Build and deploy a real-world app with MongoDB, Express, Angular, and Node (MEAN): reactive forms and custom validation."
 date: 2017-07-13 8:30
-category: Technical guide
+category: Technical guide, Angular, Angular 4
 banner:
   text: "Auth0 makes it easy to add authentication to your Angular application."
 author:
--- a/_posts/2017-07-13-real-world-angular-series-part-6.markdown
+++ b/_posts/2017-07-13-real-world-angular-series-part-6.markdown
@@ -31,7 +31,7 @@ related:
 
 ---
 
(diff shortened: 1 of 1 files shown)


_posts/2017-07-13-real-world-angular-series-part-6.markdown
```diff
@@ -0,0 +1,1610 @@
...
+
+{% include tweet_quote.html quote_text="Angular reactive forms provide granular control and support robust custom validation." %}
+
+### Event Form Requirements
+
+Let's outline the requirements for our event form. This will help us plan our logic. It should also make it clear why a reactive approach is necessary. Our event form needs the following:
+
+* Title field with simple validation
+* Location field with simple validation
+* A valid start date (ie., `1/25/2018`) at least one day in the future
```
brunokrebs: I thought it should be `i.e.` instead of `ie.`. And actually I thought this would mean "that is". I mean, I always use `i.e.` when I want to explain precisely in other words and I use `e.g.` when I want to give an example, like what you have done here.

_posts/2017-07-13-real-world-angular-series-part-6.markdown
```diff
@@ -0,0 +1,1610 @@
...
+// server/api.js
+...
+/*
+ |--------------------------------------
+ | API Routes
+ |--------------------------------------
+ */
+  ...
+  // POST a new event
+  app.post('/api/event/new', jwtCheck, adminCheck, (req, res) => {
```
kmaida: It would be enough, but for clarity, I did this with both the event and RSVP endpoints to make them easier to read / understand at a glance.

_posts/2017-07-13-real-world-angular-series-part-6.markdown
```diff
@@ -0,0 +1,1610 @@
...
+        }
+        res.send(event);
+      });
+    });
+  });
+  
+  ...
+```
+
+Only admin users should be able to add, update or delete events. This endpoint needs `jwtCheck` _and_ `adminCheck` middleware. Then we'll use the `find()` method to look for an event with the request's `title`, `location`, and `startDatetime`. If an event exists that matches all these fields, it's safe to say that we're trying to create a duplicate of an existing event and we should send an error.
```
kmaida: Changed to `middleware functions`.

_posts/2017-07-13-real-world-angular-series-part-6.markdown
kmaida: 👍

PR #893, opened 2017-06-23
Real-World Angular Series - Part 7

kmaida:
Publish: July 18

### DIFF
--- a/_posts/2017-07-18-real-world-angular-series-part-7.markdown
+++ b/_posts/2017-07-18-real-world-angular-series-part-7.markdown
@@ -401,7 +401,7 @@ The only properties we'll need for display of the event list in the My RSVPs com
 
 After handling errors for retrieving RSVPs, we can then `find()` events with an `_id` present in the `_eventIdsArr`. This is done using the [MongoDB `$in` comparison query operator](https://docs.mongodb.com/manual/reference/operator/query/in/). We only want _upcoming_ events, so we'll indicate that `startDatetime` should be [greater than or equal to (`$gte`)](https://docs.mongodb.com/manual/reference/operator/query/gte/) the current datetime. We'll then pass the `_rsvpEventsProjection` we just created to get back only the properties we need.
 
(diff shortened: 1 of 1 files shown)


_posts/2017-07-18-real-world-angular-series-part-7.markdown
```diff
@@ -0,0 +1,992 @@
...
+
+Our `_getIdp()` accessor gets the identity provider from the user's account ID. The `userProfile.sub` account IDs look something like this:
+
+```bash
+google-oauth2|23C94879435023998476321
+twitter|34B23492010786950049439
+auth0|09C3764109863877665210
+```
+
+They are strings with the identity provider followed by a pipe `|` and then a string of alphanumeric characters. In order to display the user's IdP in a friendly way, we'll `split()` on the pipe and then treat the IdP to make it more readable, if necessary.
```
kukicado: Would it be a good example to write for the google-oauth2 case to remove the "-oauth2" part of the IdP?

kmaida: I do. :) Though instead of chopping it at the hyphen and then capitalizing, there's an `if` case for this:

```
} else if (idp === 'google-oauth2') {
      idp = 'Google';
```

_posts/2017-07-18-real-world-angular-series-part-7.markdown
```diff
@@ -0,0 +1,992 @@
...
+    ...
+    this.tabSub.unsubscribe();
+  }
+
+}
+```
+
+We already have all the imports necessary to add tabs to our component class. We'll add properties for a `tabSub` subscription and `tab` string to store the name of the current tab.
+
+In `ngOnInit()`, we'll add a subscribtion to `queryParams` to set the local `tab` property to the contents of the `tab` query parameter (or `'edit'` if no parameter is available).
```
kmaida: I did that because it's specifically referring to this as a string value, rather than, say, a method name. I'm not 100% sure I did this consistently throughout all articles though... it's hard to keep track...

PR #916, opened 2017-07-05
New Extend blog post: Securing Webtasks

rwtombaugh:


elliot-w-metcalf:
securing-webtasks-p1
securing-webtasks-p2

rwtombaugh:
securing-webtasks-p1
securing-webtasks-p2

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify({ question: 'Am I secure?', answer: 'Yes!' }))
+  }
+}
+```
+
+Now we can recreate the webtask and specify the shared secret on the command line with the `--secret <key>=<value>` option. We just need to make sure we use the same key `auth-secret` that we used within our webtask:
+
+```
+wt create ./secure.js --secret auth-secret=open-sesame
```
glennblock: Instead of using the CLI, show settings secrets in the [editor](https://webtask.io/docs/issue_parameters). Mention you can also use the cli.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+      response.end();
+    } else {
+      response.writeHead(200, { 'Content-Type': 'application/json' });
+      response.end(JSON.stringify({ question: 'Am I secure?', answer: 'Yes!' }))
+    }
+  });
+}
+```
+
+Unfortunately, now that our webtask code expects a digest, it isn't so straightforward to use `curl` to send the request to the webtask. Instead, let's create a simple Node.js script that will send the request to the webtask and log the response to the console. Create a javascript file with the name `make-request.js` that has the following code:
```
glennblock: "to the webtask"  could be dropped.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+But look at that first webtask we wrote: `not-secure.js`. It was three lines of code. It was a whole lot shorter and sweeter than the final version of `secure.js` that we ended up with. Sure, the short and sweet code wasn't secure, but still, nobody would deny that it would be nice to write `not-secure.js` style webtasks and yet still have the same authorization mechanism in place as with that final version of `secure.js`. 
+
+Well, it turns out we can write short and sweet webtasks and still have robust authorization. And in the next section we'll see how it's done by using the wt-compilers feature of webtasks.
+
+## Using the Compiler Feature to Handle Authentication
+
+A wt-compiler is really just a Node.js module that is responsible for compiling a given webtask. The compiler module itself gets invoked with the webtask code as a string, so if that code happened to be coffee script that needed to be transpiled into javascript before being executed, a compiler could be written to do just that.
```
glennblock: "Webtask Compiler"

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+{"question":"Am I secure?","answer":"Yes!"}
+```
+
+As we've just shown, only callers that have access to our secret can now invoke our new secure webtask. As a first pass at securing our webtask, this isn't a bad start. Since all webtasks are hosted behind an HTTPS endpoint, the SSL/TSL connection ensures that the query-string is encrypted and so sending our secret over the wire isn't the worse thing in the world. However, we can still do better.
+
+First of all, our secret is hardcoded in our `secure.js` code. That means we have to make sure we keep our code just as secure as the webtask itself. We certainly can't make the code public on Github, or even share it with other developers we may or may not trust. Keeping secrets in our code is just an awful idea and it soon becomes untenable with even the simplest of projects.
+
+So in the next section we'll look at an important feature of webtasks that allows us to use secrets without keeping them in the code. Then in the section after that we'll look at how we can avoid sending our secret over the wire with each HTTP call.
+
+## Keeping Secrets out of Webtask Code
```
glennblock: Instead of using digest and encoding, I'd suggest showing API key auth via a header. This is the common path we recommend for securing Extend extensions, I believe we secure Auth0 hooks this way by default. It is very easy to implement (just check the header) and you don't have to show handling the body parsing yourself. Also it is easy to implement as compiler. Using API keys for this post will also give us something we can point to when  customers ask how to do API keys.'

glennblock: Another advantage of using a header, is curl supports it. This way you don't need client code at all.

_posts/2017-07-06-Securing-Webtasks.markdown
glennblock: This code completely goes away and is replaced with a curl call if you use API keys

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+  });
+}
+```
+
+This code is a complete compiler, but it really isn't doing anything interesting. It takes the webtask code, `options.script`, and passes it into the Node.js compiler, `options.nodejsCompiler`, along with a callback function that gets invoked by the Node.js compiler when the webtask function is finally compiled. It then returns that compiled webtask function via the callback. Basically, our compiler so far just compiles the code and then hands it right back to the platform to invoke it.
+
+So now let's make this compiler of ours a little more interesting:
+
+```javascript
+module.exports = function (options, callback) {
```
glennblock: Compiler will be much simpler with reading a header.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+```bash
+wt create ./secure.js --secret auth-secret=open-sesame 
+   --meta wt-compiler=https://wt-40cfb5c888bb45b0605c8156ba013930-0.run.webtask.io/compiler
+```
+
+Try running the `make-request.js` script one final time and you'll see that we're still authorized to invoke the webtask. However, if we just use a simple `curl` command we'll get an HTTP 403 response because the request doesn't include a valid digest.
+
+What's great about compilers is that they can be used with multiple webtasks that all require the same cross-cutting functionality, like authorization. And with the [Auth0 Extend](https://auth0.com/extend) product, compilers allow SaaS providers to implement product specific functionality that executes before or after their end-customer's own extensibility code. This is because the SaaS provider can specify the compiler to be used for particular webtasks, which are themselves authored by end-customers.
+
+## A Real Example: SMS with each push to a Github Repo
```
glennblock: I'd recommend making this a separate post which points back to this one. That will be valuable for people who already know secrets and a natural progression after this post. You've given the reader a great amount to digest. This is a perfect point to say, in my next post I'll show how you can use secrets for talking to external services. Also I'd use webtask.io/make again for simplicity / accelerated ramp up.

_posts/2017-07-06-Securing-Webtasks.markdown
glennblock: Recommend using digest could be yet another post in the series.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+- github
+- twilio
+- authorization
+press_release: false
+---
+# Securing Webtasks
+
+In this post I am going to show you how to secure your webtasks so that your code is only invoked by the callers you have authorized.
+
+If you aren't familiar with webtasks, it's a technology we developed here at [Auth0](https://auth0.com) for safely running code behind an HTTP endpoint without having to provision and maintain a server. As a technology, webtasks are used to support the rules feature of Auth0's flagship identity product. And just recently, we launched a new product, [Auth0 Extend](https://authom.com/extend) that heavily leverages the webtask technology to enable SaaS providers to give their customers a far superior extensibility experience than with traditional webhooks.
```
tjanczuk: I'd use this para to explain the relationship between webtask-as-a-technology to auth0-extend-as-a-product to webtask-io-as-an-instance-deployment-and-playground. It will also make the product plug more subtle.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+- extensibility
+- webtask
+- github
+- twilio
+- authorization
+press_release: false
+---
+# Securing Webtasks
+
+In this post I am going to show you how to secure your webtasks so that your code is only invoked by the callers you have authorized.
```
tjanczuk: There are many scenarios for securing webtasks: from simple keys (the Auth0 Extend pattern), to prescribed patterns (like GitHub), to web application (endpoints that show login dialogs). We can secure all (which would make for a good blog post series). But since this article focuses on GitHub, it would be worthwhile to set this context and scope up front as that single sentence suggests there is a uniform solution to all scenarios.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+If you aren't familiar with webtasks, it's a technology we developed here at [Auth0](https://auth0.com) for safely running code behind an HTTP endpoint without having to provision and maintain a server. As a technology, webtasks are used to support the rules feature of Auth0's flagship identity product. And just recently, we launched a new product, [Auth0 Extend](https://authom.com/extend) that heavily leverages the webtask technology to enable SaaS providers to give their customers a far superior extensibility experience than with traditional webhooks.
+
+If you do happen to already know about webtasks, it may be because you've previously come across [Webtask.io](https://webtask.io/), a free offering that we created to allow anyone to tinker with the webtask technology. Tinkering as we are for the purpose of this post, we'll be using Webtask.io, because it makes it easy to get hosted code up and running behind an HTTP endpoint in just a matter of minutes. 
+
+## Creating a Simple Webtask
```
tjanczuk: +1. I'd remove this entire section and either point to https://webtask.io/cli, or create a dedicated, short post 
a la "From code to endpoint in 30 seconds with webtasks", which we can later refer to from all the other posts that build on top of it.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+}
+```
+
+Now you might notice that our `secure.js` code exports a function with a different signature than the `not-secure.js` code. That's because webtasks natively support two different programming models. 
+
+Our `not-secure.js` code used the more convenient programming model in which our code was invoked with a context object and a callback function. Simply by returning our own object via the callback, we were able to send back an HTTP 200 response with the object serialized as json. This programming model is simple and intuitive to use.
+
+But sometimes we need access to more of the functionality of HTTP. That's exactly what this second native programming model of webtasks allows us to do. It provides us the HTTP server request and response objects used within Node.js. Because we want to set the HTTP status code to 403 and not include a body when the request is unauthorized, we're now using this second programming model in our `secure.js` code. 
+
+In actuality, webtasks can also be configured to support custom programming models. This is done using the wt-compiler feature, which we'll take a look at later in this post.
```
tjanczuk: `wt-compiler` is the name of the metadata property, not the feature. This is a good opportunity to settle on what we want to call it, and the use it consistently: webtask middleware, or webtask compilers?

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+However, there are also scenarios in which you wouldn't want this at all. Instead you'd want to ensure only certain services or callers were able to invoke your code. To this end, we'll consider one mechanism for securing our webtask in the next section: by using a shared secret.
+
+## Using a Shared Secret for Authorization
+
+As more and more software has moved online as a service, webhooks have grown in popularity as a way for users to respond to events from within the platform. From Slack to Github to Zendesk, webhooks are a powerful mechanism for invoking custom code and webtasks are a great fit for such a scenario. In fact, later on in this post we'll create a webtask that uses Twilio to send out an SMS message whenever a change is committed to a given Github repo.
+
+However, if you create a webtask that sends out an SMS message when triggered by Github, you wouldn't want anyone but Github to be able to invoke your webtask. In scenarios like this, the proper way to secure your webtask is with a shared secret, that is, a secret that is known by both you and the caller, but no one else.
```
tjanczuk: I'd say "one way" instead of "the proper way", since there are really many ways to accomplish this.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+To illustrate this, let's create a new webtask that only returns an HTTP 200 response if the caller sends along the secret. In your code editor, create a new javascript file named `secure.js` with the following code:
+
+```javascript
+module.exports = function(context, request, response) {
+  const query = require('url').parse(request.url, true).query
+  if (!query || query.secret != 'open-sesame') {
```
tjanczuk: query params are available off of `context.query`. So this would be `context.query.secret !== 'open-sesame'`

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+    response.writeHead(403);
+    response.end();
+  } else {
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify({ question: 'Am I secure?', answer: 'Yes!' }))
+  }
+}
+```
+
+Now you might notice that our `secure.js` code exports a function with a different signature than the `not-secure.js` code. That's because webtasks natively support two different programming models.
```
tjanczuk: Actually three not two: `(cb)`, `(ctx, cb)`, and `(ctx, req, res)`.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+Our `not-secure.js` code used the more convenient programming model in which our code was invoked with a context object and a callback function. Simply by returning our own object via the callback, we were able to send back an HTTP 200 response with the object serialized as json. This programming model is simple and intuitive to use.
+
+But sometimes we need access to more of the functionality of HTTP. That's exactly what this second native programming model of webtasks allows us to do. It provides us the HTTP server request and response objects used within Node.js. Because we want to set the HTTP status code to 403 and not include a body when the request is unauthorized, we're now using this second programming model in our `secure.js` code. 
+
+In actuality, webtasks can also be configured to support custom programming models. This is done using the wt-compiler feature, which we'll take a look at later in this post.
+
+But before we get ahead of ourselves, let's get back to creating our new secure webtask:
+
+```bash
```
tjanczuk: The problem with this is that a) you cannot copy code off of a screenshot, b) editor UI is a living thing, so not long from now these screenshots will be out of date. CLI has a much more stable interface because we try to keep it backwards compatible. It would be good however to mention webtasks can be created either through CLI  or editor and that there is functional parity between then, maybe show one screenshot of the editor to break down the content a bit.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+HTTP/1.1 403 Forbidden
+...
+```
+
+However, if we invoke the webtask with the secret provided in the query-string:
+
+```bash
+curl https://<your-webtask.io-account-domain>.run.webtask.io/secure?secret=open-sesame
```
tjanczuk: I'd consider supporting good practices by passing the secret as `Authorization: Bearer {secret}` header instead.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+In the next section, let's make another improvement to our shared secret mechanism such that the secret is also never sent over the wire. We'll do this by forcing the caller to generate a digest using our shared secret. As an added bonus, this will also ensure that the message sent to our webtask hasn't been altered.
+
+## Using A Shared Secret to Generate a Digest
+
+As mentioned earlier, all webtasks are hosted behind an HTTPS endpoint and the SSL/TSL connection ensures that the HTTP request is encrypted. Still, it's better not to send our shared secret across the wire with each request. Instead, we can require that the caller only prove it has the secret without actually sending it. We'll do this by having the caller generate a digest of the request payload using the shared secret. Within the webtask code, we'll also generate our own digest to compare with the one sent by the caller. If the two digests match, then we can be sure the caller knows the shared secret. Later in this post when we create our webtask to send an SMS message with each commit to a given repo, we'll see that Github uses this exact mechanism of generating a digest.
```
tjanczuk: Instead of `generate a digest of the request payload using the shared secret` I'd just say `sign the request payload with the shared secret using the HMAC algorithm`, and then follow through with `validate the signature` on the receiving end.

tjanczuk: It not only proves the knowledge of the secret but also ensures integrity of the payload.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+});
+
+request.write(body);
+request.end();
+```
+
+Now if we run our `make-request.js` script we'll see that we're still authorized to invoke the webtask even though we're not sending the shared secret across the wire:
+
+```bash
+node ./make-request.js
```
tjanczuk: I would still stay with curl to send a digest-protected payload. Introducing Node here adds unnecessary complexity to explaining what happens in the webtask code. You can do HAMC in bash with something as simple as

```
echo -n "value" | openssl dgst -sha1 -hmac "key"
```

_posts/2017-07-06-Securing-Webtasks.markdown
tjanczuk: +1

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+Well, it turns out we can write short and sweet webtasks and still have robust authorization. And in the next section we'll see how it's done by using the wt-compilers feature of webtasks.
+
+## Using the Compiler Feature to Handle Authentication
+
+A wt-compiler is really just a Node.js module that is responsible for compiling a given webtask. The compiler module itself gets invoked with the webtask code as a string, so if that code happened to be coffee script that needed to be transpiled into javascript before being executed, a compiler could be written to do just that.
+
+However, compilers can also be used to interject common code before or after a webtask executes, and in this post we're going to use a custom compiler to handle authorization.
+
+To get started with our own compiler, let's first consider what a default compiler module might look like:
```
tjanczuk: I'd say "no-op" or "pass-through" rather than "default".

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+    if (error) {
+      return callback(error);
+    }
+
+    return callback(null, webtaskFunc);
+  });
+}
+```
+
+This code is a complete compiler, but it really isn't doing anything interesting. It takes the webtask code, `options.script`, and passes it into the Node.js compiler, `options.nodejsCompiler`, along with a callback function that gets invoked by the Node.js compiler when the webtask function is finally compiled. It then returns that compiled webtask function via the callback. Basically, our compiler so far just compiles the code and then hands it right back to the platform to invoke it.
```
tjanczuk: The `options.nodejsCompiler` *is* the default webtask compiler.

tjanczuk: > It then returns that compiled webtask function via the callback
This re-states what the previous sentence said I think. Maybe remove?

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    return callback(null, function(context, request, response) {
+
+      var body = '';
+      request.on('data', function (data) { body += data; });
+      request.on('end', function () {
+
+        context.body = body ? JSON.parse(body) : {};
+        webtaskFunc(context, function (error, responseObject) {
```
tjanczuk: There is a quiet assumption here that the webtask signature has arity of 2 (`(ctx, cb)`). At minimum it should be asserted in code. If you want to keep it generic you'd need to do something like https://github.com/tjanczuk/wtc/blob/master/extend_context.js.

_posts/2017-07-06-Securing-Webtasks.markdown
tjanczuk: Right after introducing this no-op compiler, I'd show how to plug it in when creating a webtask before moving on to creating a more advanced version.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+          if (error) {
+            return callback(error);
+          }
+
+          if (responseObject) {
+            response.writeHead(200, { 'Content-Type': 'application/json' });
+            response.end(JSON.stringify(responseObject));
+          } else {
+            response.writeHead(200);
```
tjanczuk: To maintain parity with the behavior of `(ctx, cb)` webtasks, in the absence of response object we should also return an application/json response with an empty object in the body

```
res.writeHead(200, { 'Content-Type': 'application/json });
res.end(JSON.stringify(responseObject || {}));
```

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+        });
+      });
+    });
+  });
+}
+```
+
+This version of our compiler certainly contains more code than the previous version, but it still isn't really doing anything the webtask platform doesn't already do for us. This version of our compiler simply converts from the HTTP-level programming model of webtasks to the more convenient programming model that we used with the `not-secure.js` webtask we first wrote. 
+
+However, what's important about this version of our compiler code is how it illustrates that the compiler module actually invoke the webtask. From here, it should now be apparent how we can implement authorization at the compiler level. Since we have access to the request body in the compiler, we'll generate the digest and compare it the digest provide by the caller. If the two digests don't match, we won't even invoke the webtask code, but just return an HTTP 403 response directly. If the digests do match, we'll invoke the webtask as usual:
```
tjanczuk: ...and compare it **to** the digest **provided** by the caller...

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+  callback(null, { 
+    question: 'Am I secure?', 
+    answer: 'Yes!!! The compiler did all the authorization!' 
+  });
+}
+```
+
+Creating a webtask with the `wt` cli tool that uses a given compiler is done by including the `--meta wt-compiler=<compiler-source>` option. The `<compiler-source>` can be one of two types of values:
+
+- The name of a npm module available on the public npm repository.
```
tjanczuk: This requires that this module is also specified as a dependency with `package.json` or `--dependency` flag.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+- A url of the hosted compiler code.
+
+For the purposes of this blog, I've gone ahead and just created a webtask that returns the above compiler code. Since I haven't secured this webtask, it is publically available on the internet and you can see the code [here](https://wt-40cfb5c888bb45b0605c8156ba013930-0.run.webtask.io/compiler):
```
tjanczuk: It may be simpler to show people how to create a github gist and get a raw URL to it.

_posts/2017-07-06-Securing-Webtasks.markdown
tjanczuk: +1. I'd break it up into 2 post, we can publish them together and cross-link. By now people's heads are already exploding. The Github post can be less theoretical and more task-oriented. Ideally it should pop up when people search for "create github webhook". 

Re: CLI vs /make. How about we use CLI throughout the post so that people can copy/paste the bits, but at some in the post show an **animated gif** illustrating the end-to-end using /make, also possibly adding the webhook to github?

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+## A Real Example: SMS with each push to a Github Repo
+
+As promised, we'll finish up this blog post with a real use case in which we use webtasks to send an SMS message via Twilio each time a change is committed to a particular Github repo. Of course, you will need both a Github and a Twilio account to implement this webtask yourself, but fortunately both are free and neither require a credit card.
+
+If you don't already have a Twilio account, you can setup a free trial account [here](https://www.twilio.com/). Once you have an account you'll be able to get your `ACCOUNT SID` and `AUTH TOKEN` from the console dashboard. We'll need these when we use the Twilio npm module in our webtask. You'll also need an SMS-enabled phone number to send SMS messages from, which you can get from the console [here](https://www.twilio.com/console/phone-numbers/search). Don't worry that the console reads 'Buy a Number`, as your free trial account comes with enough credit to cover the phone number.
```
tjanczuk: Wrong closing tick after `Buy a Number`.

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+The next step is setting up Github to call the webtask we're going to create. Select any repo in your account, and click on the 'Settings' tab and then select 'Webhooks' in the left navigation bar. On the 'Webhooks' page click the 'Add Webhook' button:
+
+![Addding a Webhook](https://cdn.auth0.com/website/blog/extend/securing_webtasks/github_webhooks1.png) 
+
+Next, enter the url for the webtask we're going to create. We'll name our new webtask `github` so the url should be: `https://<your-webtask.io-account-domain>.run.webtask.io/github`. Make sure that `application/json` is selected as the content type. Lastly, enter a value for the shared secret. You'll need to use this same value for the secret when we create the webtask.
```
tjanczuk: I wonder if this would be a reasonable step to do *after* the webhook is created?

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+                      --secret TWILIO_AUTH_TOKEN=<your-auth-token>
+                      --secret TWILIO_FROM_NUMBER=<your-twilio-from-number>
+                      --secret MY_PHONE_NUMBER=<your-phone-number>
+```
+
+Now if you make a commit to your Github repo, you'll get a push notification. 
+
+However, since we aren't using a compiler and we aren't verifying the digest sent by github directly in our webtask, anyone can invoke our Github webtask. The last step is to recreate the webtask and specify the compiler and the github secret so that only authorized requests from Github will be accepted. 
+
+The compiler we'll use is very similar to the one we used earlier, but with a slight modification because Github provides the digest in a 'x-hub-signature` header instead of in the query-string. I've already created this github specific compiler and made it publically avialable as a webtask, so you can view the code [here](https://wt-40cfb5c888bb45b0605c8156ba013930-0.run.webtask.io/github-compiler).
```
tjanczuk: publically -> publicly?

tjanczuk: Again, I'd store it in some github repo folks can clone, and use raw URL to the code.

tjanczuk: Also: wrong tick at the beginning of `x-hub-signature`

_posts/2017-07-06-Securing-Webtasks.markdown
```diff
@@ -0,0 +1,483 @@
...
+
+## Recap
+
+In this post you've seen how to secure your webtasks using a shared secret to ensure only authorized callers can invoke your code. You've also learned about certain webtask features, like secrets and the wt-compiler, that make it easy to implement a shared secret solution that does all of the following:
+
+- Keeps your secrets out of your webtask code
+- Keeps your shared secret from being sent across the wire
+- Keeps your webtask code short and sweet
+
+Feel free to comment below about your experiences using webtasks. We'd especially love to hear if you've implemented a shared secret mechanism like we've discussed in this post.
```
tjanczuk: Given the CTA for this post is signup to webtask.io (https://docs.google.com/spreadsheets/d/1G_oVXC0_bhtJEIJoBnyqAPZi6unxSsi6VW2_YiE_aZo/edit#gid=0), I'd add something to that effect in the recap, e.g. [Webtask.io](https://webtask.io) is freemium deployment of [Auth0 Extend])(https://auth0.com/extend) that allows individual developers to quickly create one-off endpoints implemented in Node.js and is a convenient way of hosting webhooks.

PR #934, opened 2017-07-12
Build SaaS for change. . .not perfection

unicodeveloper:
screencapture-localhost-4000-blog-build-saas-for-change-not-perfection-1516289907522


_posts/2017-07-12-extensibility-is-the-next-big-thing-in-saas.markdown
```diff
@@ -0,0 +1,80 @@
+---
+layout: post
```
tjanczuk: According to Tony, the idea is to make this blog post part of the Auth0 Extend series. This requires the following metadata: 

```
layout: post_extend
is_extend: true
```

PR #989, opened 2017-08-15
Auth0 Extend Blog: Extend Salesforce with Node.js - Updates

NotMyself:
- [x] Add tl;dr section as a call to action
- [x] Update footer banner to be template-ized
- [x] Override footer banner content to point to auth0.com/extend/salesforce
- [ ] Get updated header image that looks more like [this](https://cdn.auth0.com/blog/office365-googleapps/logo.png)


NotMyself:
screencapture-0-0-0-0-4000-blog-extend-salesforce-with-node-1502823481204

PR #1019, opened 2017-09-13
Developing RESTful APIs with Python and Flask

brunokrebs:


_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+```
+
+Note that besides the `Transaction` class, we also defined a `TransactionSchema`. We will use the latter to deserialize and serialize instances of `Transaction` from and to JSON objects. This class inherits from another superclass called `Schema` that is defined on a package yet to be installed.
+
+```bash
+# installing marshmallow as a project dependency
+pipenv install marshmallow
+```
+
+[`marshmallow` is a popular Python package](https://marshmallow.readthedocs.io/en/latest/) for converting complex datatypes, such as objects, to and from native Python datatypes. Basically, we can use this package to validate, serialize, and deserialize data. We won't dive into validation in this article, as it will be the subject of another one. Though, as mentioned, we will use `marshmallow` to serialize and deserialize entities through our endpoints.
```
kmaida: Capitalize Marshmallow in this case because it's used to begin a sentence. If preferred, you can reword it so it's not first, e.g.:

"The popular Python package `mashmallow` converts complex..."

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+    return '<Income(name={self.description!r})>'.format(self=self)
+
+
+class IncomeSchema(TransactionSchema):
+  @post_load
+  def make_income(self, data):
+    return Income(**data)
+```
+
+The only value that this class adds for our application is that it hardcodes the type of the transaction. This type is a [Python enumerator](https://docs.python.org/3/library/enum.html), which we still have to create, that will help us filtering transactions in the future. Let's create another file, called `transaction_type.py`, inside `model` to represent this enumerator:
```
kmaida: `filtering transactions` -> "filter transactions"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+  @post_load
+  def make_expense(self, data):
+    return Expense(**data)
+```
+
+Similar to `Income`, this class hardcodes the type of the transaction, but now it passes `EXPENSE` to the superclass. What makes it different is that it forces the  `amount` passed to be negative. Therefore, no matter if the user sends a positive or a negative value, we will store it as negative to facilitate calculations.
+
+## <span id="marshmallow-serilization"></span> Serializing and Deserializing Objects with Marshmallow
+
+Having the `Transaction` superclass and its specializations properly implemented, we can now enhance our endpoints to deal with these classes. Let's replace `./cashman/index.py` contents to:
```
kmaida: `Having` -> "With"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+  expense = ExpenseSchema().load(request.get_json())
+  transactions.append(expense.data)
+  return "", 204
+
+
+if __name__ == "__main__":
+    app.run()
+```
+
+The new version that we just implemented starts by redefining the `incomes` variable into a list of `Expenses` and `Incomes`, now called `transactions`. Besides that, we have also changed the implementation of both methods that deal with incomes. On the endpoint used to retrieve incomes, we defined an instance of `IncomeSchema` to produce JSON representation of incomes. We also used [`filter`](https://docs.python.org/3/library/functions.html#filter) to extract incomes only from the `transactions` list. In the end we send the array of JSON incomes back to users.
```
kmaida: `On the endpoint used to` -> "For the endpoint used to" 

An "endpoint" is actually one end of a communication channel. Therefore, an endpoint doesn't refer to something like a function, per se, but rather, the URL used to communicate with the API. We likely wouldn't say "on" or "in" when referring to an endpoint. Consider this definition for the paragraph below too.

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+
+The endpoint responsible for accepting new incomes was also refactored. The change on this endpoint was the addition of `IncomeSchema` to load an instance of `Income` based on the JSON data sent by the user. As the `transactions` list deals with instances of `Transaction` and its subclasses, we just added the new `Income` in that list.
+
+The other two endpoints responsible for dealing with expenses, `get_expenses` and `add_expense`, are almost copies of its `income` counterparts. The differences are:
```
kmaida: `copies of its` -> "copies of their"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+# add a new income
+curl -X POST -H "Content-Type: application/json" -d '{
+    "amount": 300.0,
+    "description": "loan payment"
+}' http://localhost:5000/incomes/
+```
+
+## <span id="flask-on-docker"></span> Dockerizing Flask Applications
+
+As we are planning to eventually release our API in the cloud, we are going to create a `Dockerfile` to describe what is needed to run the application on a Docker container. We need to [install Docker on our development machine](https://docs.docker.com/engine/installation/) to test and run dockerized instances of our project. Defining a Docker recipe (`Dockerfile`) will help us running the API on different environments. That is, in the future, we will also install Docker and run our program on environments like [production](https://en.wikipedia.org/wiki/Deployment_environment#Production) and [staging](https://en.wikipedia.org/wiki/Deployment_environment#Staging).
```
kmaida: `will help us running the API` -> "will help us run the API"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+
+# Install API dependencies
+RUN pipenv install
+
+# Start app
+EXPOSE 5000
+ENTRYPOINT ["/usr/src/app/bootstrap.sh"]
+```
+
+The first item in the recipe defines that we are going to create our Docker container based on the default [Python 3 Docker image](https://hub.docker.com/_/python/). After that, we update APK and install pipenv. Having `pipenv`, we define the working directory that we will use in the image, and we copy the code needed to bootstrap and run the application. On the fourth step, we use `pipenv` to install all our Python dependencies. Lastly, we define that our image will communicate through port `5000` and that this image, when executed, needs to run the `bootstrap.sh` script to start Flask.
```
kmaida: install pipenv. Having `pipenv`, we 

You have pipenv not in backticks followed by `pipenv` in backticks for the rest of the paragraph. It might be good to add backticks to this first instance.

`On the fourth step` -> "In the fourth step"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+# run a new docker container named cashman
+docker run --name cashman \
+    -d -p 5000:5000 \
+    cashman
+
+# fetch incomes from the dockerized instance
+curl http://localhost:5000/incomes/
+```
+
+The `Dockerfile` is simple, but effective, and using it is similarly easy. With these commands and this `Dockerfile`, we can run as many instances of our API as we need with no trouble. It's just a matter of defining another port on the host, or even another host.
```
kmaida: Remove the comma after `simple,`

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+- Support for [social identity providers (like Facebook, GitHub, Twitter, etc.)](https://auth0.com/docs/identityproviders).
+- [Enterprise identity providers (Active Directory, LDAP, SAML, etc.)](https://auth0.com/enterprise).
+- Our [own database of users](https://auth0.com/docs/connections/database/mysql).
+
+[To learn the best way to secure *Python APIs* with Auth0, take a look at this tutorial](https://auth0.com/docs/quickstart/backend/python). Besides providing tutorials for backend technologies (like Python, Java, and PHP), [the *Auth0 Docs* webpage also provides tutorials for *Mobile/Native apps* and *Single-Page applications*](https://auth0.com/docs).
+
+## <span id="next-steps"></span> Next Steps
+
+In this article we learned about the basic components needed to develop a well structured Flask application. We took a look on how to use `pipenv` to manage the dependencies of our API. After that, we installed and used Flask and Marshmallow to create endpoints capable of receiving and sending JSON responses. In the end we also took a look on how to dockerize the API, which will facilitate the release of the application to the cloud.
```
kmaida: `we also took a look on how` -> "we also took a look at how"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+
+Although well structured, our API is not that useful yet. Among the things that we can improve, we are going to cover in the next article the following topics:
```
kmaida: `we are going to cover in the next article the following topics` -> "we are going to cover the following topics in the next article"

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+
+- Database persistence with SQLAlchemy.
+- Global Exception Handling.
+- Internationalization (i18n).
+- Security with JWTs.
+
```
kmaida: These are just list items, not sentences, so they shouldn't end with periods.

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+## <span id="securing-python-apis"></span> Securing Python APIs with Auth0
+
+Securing Python APIs with Auth0 is very easy and brings a lot of great features to the table. With Auth0, we only have to write a few lines of code to get:
+
+- A solid [identity management solution](https://auth0.com/user-management), including [single sign-on](https://auth0.com/docs/sso/single-sign-on).
+- [User management](https://auth0.com/docs/user-profile).
+- Support for [social identity providers (like Facebook, GitHub, Twitter, etc.)](https://auth0.com/docs/identityproviders).
+- [Enterprise identity providers (Active Directory, LDAP, SAML, etc.)](https://auth0.com/enterprise).
+- Our [own database of users](https://auth0.com/docs/connections/database/mysql).
+
```
kmaida: These are just list items, not full sentences, so they shouldn't end with periods.

_posts/2017-09-13-developing-restful-apis-with-python-and-flask.markdown
```diff
@@ -0,0 +1,540 @@
...
+```bash
+python --version
+# Python 3.6.2
+```
+
+Note that the command above might produce a different output in case we have a different Python version. What is important is that the output begins with `Python 3`, and not `Python 2`. If we get the latter, we can try issuing `python3 --version`. If this command produces the correct output, then we have to replace all commands throughout the article.
+
+### Installing Pip
+
+[Pip, which stands for Python Package Index](https://pypi.python.org/pypi/pip), is the recommended tool for installing Python packages. While the [official installation page](https://pip.pypa.io/en/stable/installing/) states that `pip` is already installed if we're using Python 2 >= `2.7.9` or Python 3 >= `3.4`, installing Python through `apt` on Ubuntu doesn't install `pip`. Therefore, let's check if we do need to install `pip` separately, or if we already have it.
```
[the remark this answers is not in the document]
brunokrebs: Well, it looks like that Pip is actually a [recursive acronym](https://en.wikipedia.org/wiki/Recursive_acronym) (I didn't know such a thing existed) for "Pip Installs Package". I'm updating the explanation.

PR #1098, opened 2017-10-28
Fix polymer package name, appears to be case sensitive

lukeoliff:
I noticed, while following this guide, that the polymer install failed with a 404 on `polymer-CLI` but worked perfectly with `polymer-cli`.

```bash
$ npm install -g polymer-CLI
npm ERR! code E404
... etc
$ npm install -g polymer-cli
npm WARN deprecated babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update! 
... etc
```

Appears to be lowercase on their guide: https://www.polymer-project.org/1.0/docs/tools/polymer-cli


kmaida:
Weird that I even put this in caps in the first place. Possibly some kind of auto-correct issue...

PR #1140, opened 2017-11-19
Auth0 Extend Blog: Securing Webtasks Part 2: Using Middleware

NotMyself:
screencapture-0-0-0-0-4000-blog-securing-webtasks-part-2-using-middleware-1511131347274


_posts/2017-11-21-securing-webtasks-part-2-using-middleware.markdown
```diff
@@ -0,0 +1,378 @@
...
+  } else {
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify({ question: 'Am I secure?', answer: 'Yes!' }));
+  }
+};
+```
+
+This code currently has two distinct responsibilities. The first is to respond with an HTTP 200 and pass along a JSON formatted object. The second is to check the query string for a secret value and ensure it matches an expected value, sending an HTTP 401 if the check fails.
+
+The code works, but it has several deficiencies. The logic needed to ensure only authorized clients can execute the task obscures what the task does. We must duplicate the authorization code in each webtask we want to be secured. Finally, the shared secret key is being passed along on the query string; this could expose the secret to any infrastructure that might be logging requests.
```
ggoodman: > this could expose the secret to any infrastructure that might be logging requests.

How does passing this information in an http header improve this situation?

_posts/2017-11-21-securing-webtasks-part-2-using-middleware.markdown
```diff
@@ -0,0 +1,378 @@
...
+
+One way of organizing shared functionality like authorization is to apply it as middleware - a self-contained bit of code that acts on a request before your webtask. Middleware can be built up as a chain of handlers in a stack; this allows them to act as plugins that can be mixed and matched to accomplish the desired result.
+
+Let's look at a simple diagram to illustrate this concept.
+
+![middleware](https://cdn.auth0.com/website/blog/extend/securing-webtasks-part-2-middleware/middleware2.png)
+
+The middleware pipeline starts on the left with an incoming HTTP request. The web framework calls the first middleware in the stack handing off the request and response.
+
+Each middleware in the stack executes and calls the next function to hand off execution down the stack. Once the last next executes, then the webtask logic executes.
```
ggoodman: I like that we're not trying to complicate things here but this isn't 100% accurate. What if we said something like this instead:

> Each middleware in the stack is executed and has the opportunity to pass control on to the next middleware or respond directly. If all middleware pass control on, then the underlying webtask will be executed.

_posts/2017-11-21-securing-webtasks-part-2-using-middleware.markdown
```diff
@@ -0,0 +1,378 @@
...
+```javascript
+module.exports = function() {
+  return function middlware(request, response, next) {
+    // logic to be executed before the webtask
+    next(); //call the next middleware
+  }
+}
+```
+
+This code exports a function that when called returns a middleware function that takes the raw HTTP request & response and a next function.
```
ggoodman:
```diff
- This code exports a function that when called returns a middleware function that takes the raw HTTP request & response and a next function.
+ This code exports a function that, when called, returns a middleware function that takes the raw HTTP request & response and a next function. If this seems familiar, it is probably because this is the same pattern is used by most express middleware.
```

_posts/2017-11-21-securing-webtasks-part-2-using-middleware.markdown
NotMyself: This was a concern that @tjanczuk brought up. He said that passing the token via query string exposed it to a lot of infrastructure that logs request urls automatically. Passing via headers avoids that at least.

_posts/2017-11-21-securing-webtasks-part-2-using-middleware.markdown
NotMyself: Grammarly says you are wrong. 😁

PR #1178, opened 2017-11-27
Replaced in-article aside with generic React aside

kmaida:
@TODO: aside must be verified (works on React 16)


kmaida:
@brunokrebs to just quickly verify that the generic React aside works with React 16 (it should).

kmaida:
Merged into a master migration branch.

PR #1217, opened 2017-12-04
Update Angular 2 Authentication tutorial

kmaida:
* Updated content to latest version of Angular CLI and Angular v5.
* Updated wording and language and instructions to match the repo updates.
* Fixed typos and added instructions where some were missing (e.g., `auth0-js` was never installed)
* Made code changes to allow lettable operators, better tree-shaking, use observables instead of promises, etc.
* Used the CLI to generate components; before, CLI was only used to create and serve the project.


kmaida:
Need input from @ggonzaleza regarding use of post filename, date, and updated metadata.

kmaida:
Confirmed that as long as the URL stays the same, updating the date does not affect SEO. 👍

PR #1253, opened 2017-12-17
Developing NPM Packages

brunokrebs:


_posts/2017-12-17-developing-npm-packages.markdown
```diff
@@ -0,0 +1,721 @@
...
+  { input: undefined, expectedResult: null, description: 'should return null when undefined is passed' },
+  // ... the initial samples stay untouched
+];
+
+// ... describe method stays untouched
+```
+
+If we ask Istanbul now (`npm run test`), we will see that we managed to add enough scenarios to cover all our source lines of code. This is not a proof that our package contains no bug, but enough to make us confident to publish its initial version. So let's do it.
+
+[Publishing NPM packages looks like a very simple process](https://docs.npmjs.com/getting-started/publishing-npm-packages). As described in the official documentation, all we need is to [create a user](https://docs.npmjs.com/getting-started/publishing-npm-packages#creating-a-user) (if we still don't have one) and then issue `npm publish` in the project root, right? Well, not so fast. Indeed is not hard to publish a NPM package, but we always want to distribute an ES5 version of our package for maximum compatibility. We could leave this as a manual process (that is, expect the developer to run `npm run build` before publishing a new version), but this is too error-prone.
```
unicodeveloper: Indeed is not hard.... => Indeed, it is not hard...

PR #1342, opened 2018-01-29
Troubleshooting Webtasks: Using the Editor

NotMyself:
Current Word Count: 1665

screencapture-localhost-4000-blog-troubleshooting-webtasks-using-the-editor-2018-03-12-10_13_56


cfjedimaster:
This part: "Auth0 Hooks uses the editor to offer templates tailored to the customization point. During the Client Credentials Exchange event, webtasks give access to the calling client, the requested scope, and audience that allows for customization of the generated access token."

As someone new to WT, I have no idea what this is. Do you think most of the audience would? It kinda explains itself based on the name, but it definitely threw me.

cfjedimaster:
"CMD-L" Consider CMD/CTRL-L please. :)

cfjedimaster:
Meta comments:
Maybe consider screen shots in case a user doesn't want to do any work?
For `wt serve`, maybe a quick warning that the CLI doesn't support automatic reloading, so you need to ctrl-c, rerun the code on changes.

cfjedimaster:
Possible typo: "We can easily write our messages " - did you mean "out", not "our"?

cfjedimaster:
typo: "Clicking it will cause the current webtask code to saved a"  should be "to be saved"

cfjedimaster:
"This code uses the requestjs node" - is it called requestjs? I thought it was just the 'request' module.

cfjedimaster:
Note on last comment - you used requestjs more than once, so if you do change it, change it every time.

PR #1373, opened 2018-02-21
Using Python, Flask, and Angular to Build Modern Apps - Part 1

brunokrebs:


_posts/2018-02-21-using-python-flask-and-angular-to-build-modern-apps-part-1.markdown
```diff
@@ -0,0 +1,620 @@
...
+touch src/__init__.py
+
+# create file to mark entities as a module
+touch src/entities/__init__.py
+
+# create file to hold some boilerplate code
+touch src/entities/entity.py
+```
+
+The first two `touch` commands above simply create empty `__init__.py` files to mark both directories as Python modules. The last `touch` command creates the file that will hold a class called `Entity`. You will use this class as the superclass to all your entities. This will be useful to avoid having to repeat some boilerplate code to connect to the database and to define some common properties (e.g. `id` and `created_at`):
```
lukeoliff: database and to define
->
database, and to define

?

PR #1394, opened 2018-03-02
Common threats in web application security

lukeoliff:


_posts/2018-03-01-common-threats-in-web-application-security.markdown
```diff
@@ -0,0 +1,241 @@
...
+
+  This is basically a boolean value that determines whether XSS filtering is enabled. Change it to `0` to disable it. Without the optional `mode` or `report` directives, the browser will just sanitize the page, even removing the affected parts.
+  
+- `mode=block` _Optional_
+
+  Enables XSS filtering. Rather than sanitizing the page, the browser will prevent rendering of the page if an attack is detected.
+  
+- `report=<report url>`
+
+  Enables XSS filtering. If a cross-site scripting attack is detected, the browser will sanitize the page and report the violation.
```
sebadoom: It may be worth mentioning that this feature is not supported by Firefox and that it is redundant with the Content-Security-Policy directive in place. See here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

_posts/2018-03-01-common-threats-in-web-application-security.markdown
```diff
@@ -0,0 +1,241 @@
...
+
+However, if we do not wish to allow *any* external sites to execute scripts on our web application, we could simply include the following:
+
+```js
+function requestHandler(req, res) {
+	res.setHeader( 'Content-Security-Policy', "script-src 'self'" );
+}
+```
+
+Note the `script-src` directive here, that we have set to `self`, therefore only allowing scripts from within our own domain. Content Security Policies are both excellent and very powerful, but must be used cautiously. Just as with HSTS, incorrect configuration could cause unforeseen issues or missing content.
```
sebadoom: Note that this disables inline scripts, which is a big win for security. Perhaps it is worth mentioning that here.

_posts/2018-03-01-common-threats-in-web-application-security.markdown
lukeoliff: When I reviewed the last version of this post, I actually recommended providing a compatibility table of our own (like Moz's). Not sure why I didn't follow my own advice! :) 👍

_posts/2018-03-01-common-threats-in-web-application-security.markdown
lukeoliff: Ok cool

PR #1419, opened 2018-03-15
React Router 4: A Practical Introduction

unicodeveloper:


_posts/2018-03-05-react-router-4-practical-tutorial.markdown
```diff
@@ -0,0 +1,810 @@
...
+* Nested Routing and URL Parameters
+* Route Protection and Authentication
+* Link Component Customization
+* Handling Non-existent Routes
+* SideBar Rendering
+
+
+### Basic Routing
+
+There are two types of Router components that you can use in your React web application. The `BrowserRouter` and `HashRouter`. The former gives you a URL without the `#`, while the latter gives you a URL wit the `#`.
```
lukeoliff: while the latter gives you a URL wit the
->
while the latter gives you a URL with the

_posts/2018-03-05-react-router-4-practical-tutorial.markdown
```diff
@@ -0,0 +1,810 @@
...
+      <div>
+        <p>You must log in to view the page at {from.pathname}</p>
+        <button onClick={this.login}>Log in</button>
+      </div>
+    );
+  }
+}
+```
+
+We have modified the Login Component to be able to have a `login` function and also redirect back to the route that the user was trying to log onto when the user was denied access. This should be typical behaviour of your routing system else users will always be redirected to a particular page rather than where they came from!
```
lukeoliff: behaviour -> behavior

we default to US spelling I think :(

_posts/2018-03-05-react-router-4-practical-tutorial.markdown
```diff
@@ -0,0 +1,810 @@
...
+            <Route
+              key={route.path}
+              path={route.path}
+              exact={route.exact}
+              component={route.leftbar}
+            />
+          ))}
+        </div>
+
+        <div style={{ flex: 1, padding: '20px' }}>
```
brunokrebs: I'm almost sure this won't render on our blog: https://github.com/auth0/blog/blob/master/post-cheat-sheet.markdown#html

@unicodeveloper @lukeoliff

_posts/2018-03-05-react-router-4-practical-tutorial.markdown
```diff
@@ -0,0 +1,810 @@
...
+  )
+}
+```
+
+The `<Route/>` component had an `onEnter` prop that accepts a method that allows entry or refusal to a URL location based on a user's authentication status. Now, it's different in React Router 4.
+
+Let's build out three components, `Public`, `Private`, and `Login`.
+
+_App.js_
+import React, { Component } from 'react';
```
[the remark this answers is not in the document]
brunokrebs: Actually, inside this type of block (as it contains `{{ }}`): https://github.com/auth0/blog/blob/master/post-cheat-sheet.markdown#html

PR #1440, opened 2018-03-28
Token Security: Stop Putting Your Users At Risk

lukeoliff:


_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+- breaches
+related:
+- 2018-02-27-8-ways-to-avoid-healthcare-breaches
+- 2018-02-14-what-is-data-security
+- 2018-03-22-cambridge-analytica-and-facebook§t
+---
+
+**TL;DR:** We explore keeping your tokens and your user’s data safe. `localStorage` has been used as an example of how to store access tokens, especially in single-page applications (SPAs). Auth0 has an alternative to storing a token for a persistent user session, so your returning users can stay logged in without putting them at unnecessary risk.
+
+## Storing tokens on the frontend—don’t do it
```
kmaida: General recommendation throughout the article: fewer em dashes.

* http://www.slate.com/articles/life/the_good_word/2011/05/the_caseplease_hear_me_outagainst_the_em_dash.html
* https://www.quickanddirtytips.com/education/grammar/when-to-use-and-not-use-an-em-dash
* https://www.prdaily.com/Main/Articles/Writers_beware_Em_dashes_are_overused_and_misunder_5368.aspx

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Access tokens are a credential that can be used by a client to access a service, like an application programming interface (API). Like any credential—passwords for instance—they should be kept safe and secure.
+
+{% include tweet_quote.html quote_text="Are you putting your users at unnecessary risk? If you store access tokens in your frontend app, you might be!" %}
+
+You should not be storing anything as sensitive as credentials anywhere they can be accessed by a third party and unfortunately storing it in your frontend application is asking for trouble.
```
kmaida: This is a run-on sentence with a bit of problematic grammar. I recommend splitting this into two sentences, or breaking it up with a comma. `storing it in` -> "it" is referring to credentials here, which are plural in the beginning of the sentence, but then are referred to as singular. It makes the sentence sound a bit strange.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Attackers are actively looking for places you might be exposing sensitive data and they’re using more elaborate methods all the time. The last thing you need to do is make it easy for them.
+
+A common place for storing tokens on the frontend is JavaScript’s [`Web Storage API`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)—which should be considered insecure. Better known by their function names, `localStorage` and `sessionStorage`.
```
kmaida: "Better known by their function names, `localStorage` and `sessionStorage`." 

This is an incomplete sentence.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+The benefit of the `Web Storage API` is that it’s purely JavaScript and allows you to store up to 5MB of data. The data can be stored as string data only, therefore it has no logic or state. But it can be serialized along with its context (i.e. as JSON object). With `localStorage`, storage persists on that domain until such time as it’s deleted by the user or the site. This is unlike `sessionStorage` only in that `sessionStorage` persists data limited to the length of that browsing session.
+
+The problem with the `Web Storage API` is that it assumes anyone who can access it, also has permission to be accessing it. This is because it’s only accessible by JavaScript and based on the current origin or domain. But, this means it’s vulnerable to XSS attacks. Especially so because it’s stored in such a readily available format. It’s easier to use, but easier to abuse. For example, running the JavaScript below in your browser console will loop over every item on that site's `localStorage` and output the keys **and the values**.
```
kmaida: `anyone who can access it, also has` -> remove the comma, and I'd recommend adding emphasis to specific words to highlight the comparison this sentence is making. E.g.:

"The problem with the `Web Storage API` is that it assumes anyone who _can_ access it also has _permission_ to access it."

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+```js
+Object.keys(localStorage).forEach(function(key){
+   console.log(`${key}: ${localStorage.getItem(key)}`);
+});
+```
+
+> “So what? That doesn’t help an attacker get your tokens. That's just what I can run on my console!”—You (maybe).
```
kmaida: I kind of love this.

Missing a space though between the closing quotation and `—You`, and don't need a period.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Let's speak theoretically. Imagine for a moment that an attacker can place some JavaScript on your site. Replace the `console.log()` above with an [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) script and **off goes your users data to some other server**.
```
kmaida: `your users data` -> "your users' data"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+What if you’re using [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to stop anything but your trusted vendor's scripts from running on your application? What if you’re using [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to ensure even those vendors scripts can’t open up vulnerabilities on your application? Those **are** pretty good ways to protect your application! ***Those security steps should be taken anyway!*** But can you be 100% certain it’s safe? Avoiding storing tokens on the frontend is the best way.
```
kmaida: `even those vendors scripts can’t` -> "vendor scripts"

`Avoiding storing tokens on the frontend is the best way.` This is somewhat problematic because SPAs use the token to make API requests. This sentence makes it sound like the token shouldn't be stored _anywhere_ in the front end, which means we would never be able to use it for anything either, unless we made a request for a new token every time we wanted to make an API request. We do still need to store it in the app so we can make requests, so let's clarify this in reference to browser storage.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+**Remember, having someone’s access token can be as good as having their username and password, and you wouldn’t ever risk exposing your username and password.**
+
+{% include tweet_quote.html quote_text="Having someone’s access token can be as good as having their username and password, and you wouldn’t ever risk exposing your username and password" %}
+
+## The complexity of web applications
+
+In the 1990’s world of basic web development, you had monolithic applications that handled user management, stored sensitive data, processed requests and returned some HTML to our screen. Those applications acted as both our server and our client application, all rolled into one.
```
kmaida: `In the 1990’s world` -> No apostrophe is necessary here.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+In the years that followed, developers made that HTML prettier and prettier, and more interactive. But they essentially still had one application which kept our user's credentials and session secure—for the most part—in the backend and still returned HTML to the frontend. Our monolithic application is **one single application**, and its just sending HTML down to our browser, which is more or less just a dumb client that lets us click on our new pretty buttons and links.
```
kmaida: `kept our user's credentials` -> "kept our users' credentials" (possessive of a plural)

`and its just sending HTML` -> "and it's just" (contraction)

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+At this point, the browser is not actively involved in any logic. All you need to do with each request is to identify the user is who they were at the start of their browsing session. In such a case you would use cookies, because that's what they were designed for, a single party that has no need to expose anything with any potential sensitivity.
```
kmaida: `they were designed for, a single` -> "they were designed for: a single" (change comma to a colon)

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Roll on two decades of user experience improvements and technological advancements and you land smack bang in the world of single-page applications (SPAs). An SPA is essentially software that runs in the browser. When you load up a new website, you download very thin code that comes with all the necessaries to be able to communicate with a server.
+
+The server remains essentially the same as our old 1990's application, but it’s learned some tricks. It has had an upgrade. It only talks in computer language now—HTTP requests sending and receiving JSON payloads. It’s now called an API and it doesn't return HTML anymore. Instead, it leaves the HTML up to the SPA.
```
kmaida: `1990's` -> "1990s"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Web tokens were designed for cases like this, in cases where there are *two or more applications*, such as our API and our SPA. It allows the SPA to talk to the API, where you can identify the authorized user. If you don't protect your tokens with all the responsibility you give other credentials and then the token it’s accessed by a bad actor, they could access your API on behalf of that your user.
```
kmaida: The first sentence feels awkward here: `or cases like this, in cases where` I'd recommend changing one use of the word "cases" to a synonym.

`the token it’s accessed` -> "the token is accessed" (typo)

`on behalf of that your user` -> "on behalf of your user" OR "on behalf of that user"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+**Now imagine that the API is a bank, and the SPA is a user's online banking software.**
+
+## Auth0’s old guides—they store tokens in localStorage!
+
+We’re way ahead of you! We’re working hard to update our existing guides.
```
kmaida: `We’re way ahead of you!` I wouldn't say this, because we're actually behind in updating our content. I'd just acknowledge this without making a statement with this type of language. Maybe something like, "We know, and we're working hard to update our existing guides!"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+A lot of people will still go ahead and use `localStorage` assuming they’ll never be vulnerable to XSS attacks, having taken the appropriate steps to protect against it.
```
kmaida: `protect against it` -> "protect against them" (`XSS attacks` are plural, but then referred to as singular)

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+That’s ok, but it’s only responsible for **Auth0 to teach you the most secure way**.
```
kmaida: `ok` -> "okay"

I'd rephrase the last part of this sentence too. It sounds kind of strange with the word `only` in it.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Until recently, we believed that with simple and common steps to secure your SPAs against XSS and other attack vectors, it was okay to use `localStorage` in guides because it made our guides very simple to follow. The reality is that the only way to build secure applications is to do absolutely **everything** you can to secure them, because ***attackers are always one step ahead***.
```
kmaida: I would focus on the pros for using local storage here, rather than just saying we thought it'd be the easiest thing for users to do in a guide. This whole paragraph currently sounds like we were _advocating for not_ doing everything to make apps secure, and we only just now decided that we should tell people how to do it right.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+It’s not just Auth0 though, [more](https://www.rdegges.com/2018/please-stop-using-local-storage/) and [more](https://www.owasp.org/index.php/JSON_Web_Token_(JWT)_Cheat_Sheet_for_Java#Token_storage_on_client_side) developers and groups are putting up red flags on token storage.
```
kmaida: Change link text to something more descriptive for better SEO and accessibility. One way to do this is to simply divide and link the whole second part of the sentence instead of just the words `more` and `more`

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+So we’ve discussed the risk of storing tokens with the `Web Storage API` and how you might mitigate the risk if you're determined to do it that way. But as I said, that’s not enough. I need to show you the right way an SPA can authenticate and re-authenticate a returning user without that token appearing from `localStorage`.
+
+The [`auth0-js` SDK](https://auth0.com/docs/libraries/auth0js) `checkSession()` method allows you to acquire a new token from Auth0 for a user who is already authenticated against Auth0 for your domain. The method accepts any valid OAuth2 parameters that would normally be sent to our `authorize()` method, which is used to initiate authentication in the first place. This allows you to re-authenticate returning users without any dramatic re-engineering.
+
+The Auth0 guides would normally include an authentication utility script to provide some much-needed functions for authentication on an SPAs. So below you'll find an example using the `auth0-js` SDK. **This example isn't working code and shouldn't be expected to run**, but it gives you a good idea of what methods you should be looking to use in order to avoid storing tokens with the `Web Storage API`.
```
kmaida: `So below you'll find` -> "Below you'll find" (avoid starting sentences with conjunctions)

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+Our utility script needs our `auth0-js` SDK loaded and then it needs to configure it. You might load it using a loader like `CommonJS` or `RequireJS`, or just use it from the CDN using traditional `script` tags.
```
kmaida: `You might load it using` -> "You might load it from Node modules using" 

Add links to CommonJS and RequireJS.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+So here is the start of the utilily script, that uses the `auth0-js` SDK.
```
kmaida: Remove "So"

`utilily` -> "utility" (typo)

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+        handleLogin(errors, authResult) // handle and store logged in user
+      );
+    }
+  }
+  
+  handleCallback();
+});
+```
+
+Previously, to handle writing and removing user access from the utility script, you would have two functions like the ones below. One to add the info and one to remove it. The callback will parse the authorization and write the token to the SPA.
```
kmaida: `below. One to add the` -> "below: one to add the"

`The callback will parse` -> you switched tenses here; the beginning of this paragraph says `Previously, you would have`

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+Before, the example used `localStorage`. But this time, we can achieve the same result without exposing the token to the `Web Storage API`.
+
+```js
+// authutil.js
+window.addEventListener('load', function() {
+  // ...
+  
+  function handleLogin(errors, authResult) {
+    localStorage.setItem('expires', authResult.expiresIn); // magic with expires time
+    var accessToken = authResult.accessToken; // store our token in the DOM
```
kmaida: We're not storing the token in the `DOM`; we're storing it in app memory.

This goes against some statements made earlier in the article about _not storing the token_. We need to explain why storing in app memory is safer than storing in local storage. That could be done in the sections above.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  function handleLogout() {
+    localStorage.removeItem('expires');
+    var accessToken = undefined;
+  }
+});
+```
+
+Using this setup as it is, a user on your SPA application would remain logged in until they re-request the SPA from the server. So, something like a page refresh or browsing around without using the SPA's router will likely log them out.
```
kmaida: Remove `So,`

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+  handleReturningUser();
+});
+```
+
+At this point you might be wondering how `checkSession()` works, right?
+
+`checkSession()` works by creating an invisible iframe which makes a request to `the-application-domain.auth0.com/authorize`. What this request does is return a new `token` to the application.
+
+It returns tokens if the user is still logged in to the authentication server. The authentication server uses normal `web sessions` to know whether the user is authenticated or not. Just like going back to Facebook a day later and it logging users back in. They can still get a `token` from the authentication server while their session still exists.
```
kmaida: `authenticated or not. Just like` -> "authenticated or not, just like"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+That session will expire eventually, according to authorization server configuration. If the session has expired, the application will receive an `interaction_required` error in the response, which is when it’s time to log in again!
+
+## Conclusion
+
+Access tokens are to an application what usernames and passwords are to a login box, representing a user until expiry.
+
+Unnecessary risks with access tokens such as storing them in `localStorage` and storing them incorrectly in a Cookie leaves them vulnerable to many attack vectors. While these attacks can be defended against, it’s only responsible of us to teach you the most secure way to handle authentication.
```
kmaida: `in a Cookie` -> "in a cookie"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+That is why we're recommending that access tokens aren't stored with the `Web Storage API`, and if they're stored in a Cookie then you should apply `HttpOnly` and `SameSite` directives.
```
kmaida: `in a Cookie` -> "in a cookie"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: I'd hyphen `front-end`.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: lol

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: I'd rephrase this in the following way:

Until recently, we presented simple and common steps to secure your SPAs against XSS and other attack vectors. This was done as a way to make our guides simple to understand and follow, which is one of the strengths of the Web Storage API. The reality is that the only way to build secure applications is to do absolutely **everything** you can to secure them, because ***attackers are always one step ahead***.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+That’s ok, but it’s only responsible for **Auth0 to teach you the most secure way**.
+
+Until recently, we believed that with simple and common steps to secure your SPAs against XSS and other attack vectors, it was okay to use `localStorage` in guides because it made our guides very simple to follow. The reality is that the only way to build secure applications is to do absolutely **everything** you can to secure them, because ***attackers are always one step ahead***.
+
+It’s not just Auth0 though, [more](https://www.rdegges.com/2018/please-stop-using-local-storage/) and [more](https://www.owasp.org/index.php/JSON_Web_Token_(JWT)_Cheat_Sheet_for_Java#Token_storage_on_client_side) developers and groups are putting up red flags on token storage.
+
+## Your returning users can still re-authenticate
+
+So we’ve discussed the risk of storing tokens with the `Web Storage API` and how you might mitigate the risk if you're determined to do it that way. But as I said, that’s not enough. I need to show you the right way an SPA can authenticate and re-authenticate a returning user without that token appearing from `localStorage`.
```
sebadoom: I -> We

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+
+The [`auth0-js` SDK](https://auth0.com/docs/libraries/auth0js) `checkSession()` method allows you to acquire a new token from Auth0 for a user who is already authenticated against Auth0 for your domain. The method accepts any valid OAuth2 parameters that would normally be sent to our `authorize()` method, which is used to initiate authentication in the first place. This allows you to re-authenticate returning users without any dramatic re-engineering.
```
sebadoom: OAuth2 -> OAuth 2.0

sebadoom: I'd add:

Note that users won't be presented with a login screen unless required. `checkSession` will get a new token transparently when certain conditions are met.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: Remove the `CommonJS` and `RequireJS` sentence, it is confusing. "CommonJS" is not a loader, for instance.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: Add a "warning" sentence before the example: "This is an example of how you should NOT do things", or something like that.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: Although I can understand what you are trying to show here, this will probably confuse some JS developers. The use of `var` makes the lifetime of the variable scoped to the function. You probably want to show that you should keep it in a global or something like that. For instance, by putting the `var accessToken` declaration at the top level (i.e. outside of the `addEventListener` function).

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+  // ...
+  
+  function handleLogin(errors, authResult) {
+    localStorage.setItem('expires', authResult.expiresIn); // magic with expires time
+    var accessToken = authResult.accessToken; // store our token in the DOM
+  }
+
+  function handleLogout() {
+    localStorage.removeItem('expires');
+    var accessToken = undefined;
```
sebadoom: This does not do what you think it does. The `accessToken` variable was already undefined at this point.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  handleReturningUser();
+});
+```
+
+At this point you might be wondering how `checkSession()` works, right?
+
+`checkSession()` works by creating an invisible iframe which makes a request to `the-application-domain.auth0.com/authorize`. What this request does is return a new `token` to the application.
```
sebadoom: Note that the use of the term `application` as part of the domain is not correct per the terms we use at Auth0. It's the `tenants` or `users` domain. A single tenant can have multiple apps.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: Cookie -> cookie

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
sebadoom: And the `Secure` directive, and possibly consider the `__Secure` and `__Host` prefixes.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: Done! I hate em dashes now. I never get them right. Haha

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,209 @@
...
+- jwt
+- auth
+- breaches
+related:
+- 2018-02-27-8-ways-to-avoid-healthcare-breaches
+- 2018-02-14-what-is-data-security
+- 2018-03-22-cambridge-analytica-and-facebook§t
+---
+
+**TL;DR:** We explore keeping your tokens and your user’s data safe. `localStorage` has been used as an example of how to store access tokens, especially in single-page applications (SPAs). Auth0 has an alternative to storing a token for a persistent user session, so your returning users can stay logged in without putting them at unnecessary risk.
```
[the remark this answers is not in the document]
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: thanks, better than my edit as well!

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: i've removed the "how not to do it" example altogether

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: good call, also added a link to my blog post where i detail their use. thanks!

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Storing tokens on the front-end, don’t do it
+
+Access tokens are a credential that can be used by a client to access a service, like an application programming interface (API). Like any credential, passwords for instance, they should be kept safe and secure.
+
+{% include tweet_quote.html quote_text="Are you putting your users at unnecessary risk? If you store access tokens in your front-end app, you might be!" %}
+
+You should not be storing anything as sensitive as credentials anywhere they can be accessed by a third party. Storing them your front-end application is asking for trouble. Attackers are actively looking for places you might be exposing sensitive data and they’re using more elaborate methods all the time. The last thing you need to do is make it easy for them.
```
unicodeveloper: > "Storing them your front-end application is asking for trouble."

Something is missing here. "is" ?

lukeoliff: Thanks! fixed!

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+The server remains essentially the same as our old 1990s application, but it’s learned some tricks. It has had an upgrade. It only talks in computer language now, HTTP requests sending and receiving JSON payloads. It’s now called an API and it doesn't return HTML anymore. Instead, it leaves the HTML up to the SPA.
+
+Web tokens were designed for use in cases where there are *two or more applications*, such as our API and our SPA. It allows the SPA to talk to the API, where you can identify the authorized user. If you don't protect your tokens with all the responsibility you give other credentials and then the token is accessed by a bad actor, they could access your API on behalf of that user.
+
+**Now imagine that the API is a bank, and the SPA is a user's online banking software.**
+
+## Auth0’s old guides, they store tokens in localStorage!
```
sebadoom: Are we sure this is the tone we want to use for this issue? Sounds like admittance of guilt, which is certainly not the case. We are offering our view on a more secure solution.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+We know and we’re working really hard to update our existing guides.
```
sebadoom: Same as above.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Storing tokens on the front-end, don’t do it
+
+Access tokens are a credential that can be used by a client to access a service, like an application programming interface (API). Like any credential, passwords for instance, they should be kept safe and secure.
+
+{% include tweet_quote.html quote_text="Are you putting your users at unnecessary risk? If you store access tokens in your front-end app, you might be!" %}
+
+You should not be storing anything as sensitive as credentials anywhere they can be accessed by a third party. Storing them in your front-end application is asking for trouble. Attackers are actively looking for places you might be exposing sensitive data and they’re using more elaborate methods all the time. The last thing you need to do is make it easy for them.
+
+A common place for storing tokens on the front-end is JavaScript’s [`Web Storage API`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), which should be considered insecure.
```
sjudson: The API shouldn't be considered insecure, the practice of storing credentials in `localStorage` should be. Also the `Web Storage API` isn't part of Javascript, it's part of the browser standards (e.g.: https://www.w3.org/TR/webstorage/).

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+The benefit of the `Web Storage API` is that it’s purely JavaScript and allows you to store up to 5MB of data. The data can be stored as string data only, therefore it has no logic or state. But it can be serialized along with its context (i.e. as JSON object). With `localStorage`, storage persists on that domain until such time as it’s deleted by the user or the site. This is unlike `sessionStorage` only in that `sessionStorage` persists data limited to the length of that browsing session.
```
sjudson: > is that it’s purely JavaScript

It can be accessed from Javascript, it's not Javascript itself, it's part of the browser.

sjudson: > can be stored as string data only, therefore it has no logic or state

Why is this relevant? It's storage, and therefore stores data, not logic.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+The problem with the `Web Storage API` is that it assumes anyone who _can_ access it, also has _permission_ to access it. This is because it’s only accessible by JavaScript and based on the current origin or domain. But, this means it’s vulnerable to XSS attacks. Especially so because it’s stored in such a readily available format. It’s easier to use, but easier to abuse. For example, running the JavaScript below in your browser console will loop over every item on that site's `localStorage` and output the keys **and the values**.
```
sjudson: >  it assumes anyone who _can_ access it, also has _permission_ to access it. This is because it’s only accessible by JavaScript and based on the current origin or domain.

This line comes across odd. I'd just say that the only access control is that you must be from the same origin.

sjudson: >  But, this means it’s vulnerable to XSS attacks. Especially so because it’s stored in such a readily available format.

Not sure what the format has to do with it. There are also other types of exfiltration attacks than just XSS, though it is the most important one.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+Object.keys(localStorage).forEach(function(key){
+   console.log(`${key}: ${localStorage.getItem(key)}`);
+});
+```
+
+> “So what? That doesn’t help an attacker get your tokens. That's just what I can run on my console!” —You (maybe)
+
+Let's speak theoretically. Imagine for a moment that an attacker can place some JavaScript on your site. Replace the `console.log()` above with an [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) script and **off goes your users' data to some other server**.
+
+What if you’re using [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to stop anything but your trusted vendor's scripts from running on your application? What if you’re using [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to ensure even those vendor scripts can’t open up vulnerabilities on your application? Those **are** pretty good ways to protect your application! ***Those security steps should be taken anyway!*** But can you be 100% certain it’s safe? Avoiding storing tokens with the `Web Storage API` is the best way.
```
sjudson: This is all assuming that XSS is the only way things can be exfiltrated out of the browser. E.g., Meltdown allowed malicious JS to read into other segments of browser process memory. Cloudbleed dumped tokens into crawler memory. This is all good, but I'd focus that there are general security concerns around exposing tokens to the large variety of attack vectors in the browser. So that e.g., even if CSP does save you from XSS, that's still not enough.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+```js
+Object.keys(localStorage).forEach(function(key){
+   console.log(`${key}: ${localStorage.getItem(key)}`);
+});
+```
+
+> “So what? That doesn’t help an attacker get your tokens. That's just what I can run on my console!” —You (maybe)
+
+Let's speak theoretically. Imagine for a moment that an attacker can place some JavaScript on your site. Replace the `console.log()` above with an [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) script and **off goes your users' data to some other server**.
```
sjudson: >  Imagine for a moment that an attacker can place some JavaScript on your site.

Note that this frequently isn't hard w/ things like advertising networks and embeddings.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+What if you’re using [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to stop anything but your trusted vendor's scripts from running on your application? What if you’re using [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to ensure even those vendor scripts can’t open up vulnerabilities on your application? Those **are** pretty good ways to protect your application! ***Those security steps should be taken anyway!*** But can you be 100% certain it’s safe? Avoiding storing tokens with the `Web Storage API` is the best way.
+
+**Remember, having someone’s access token can be as good as having their username and password, and you wouldn’t ever risk exposing your username and password.**
```
sjudson: "But hey, I put my username and password into the browser so that must be okay?"

Do you see what the difference is?

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+At this point, the browser is not actively involved in any logic. All you need to do with each request is to identify the user is who they were at the start of their browsing session. In such a case you would use cookies, because that's what they were designed for: a single party that has no need to expose anything with any potential sensitivity.
+
+Roll on two decades of user experience improvements and technological advancements and you land smack bang in the world of single-page applications (SPAs). An SPA is essentially software that runs in the browser. When you load up a new website, you download very thin code that comes with all the necessaries to be able to communicate with a server.
+
+The server remains essentially the same as our old 1990s application, but it’s learned some tricks. It has had an upgrade. It only talks in computer language now, HTTP requests sending and receiving JSON payloads. It’s now called an API and it doesn't return HTML anymore. Instead, it leaves the HTML up to the SPA.
```
sjudson: >  It only talks in computer language now

Umm, what language was it talking in before?

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+In the years that followed, developers made that HTML prettier and prettier, and more interactive. But they essentially still had one application which kept our users' credentials and session secure, for the most part, in the backend and still returned HTML to the front-end. Our monolithic application is **one single application**, and it's just sending HTML down to our browser, which is more or less just a dumb client that lets us click on our new pretty buttons and links.
+
+At this point, the browser is not actively involved in any logic. All you need to do with each request is to identify the user is who they were at the start of their browsing session. In such a case you would use cookies, because that's what they were designed for: a single party that has no need to expose anything with any potential sensitivity.
+
+Roll on two decades of user experience improvements and technological advancements and you land smack bang in the world of single-page applications (SPAs). An SPA is essentially software that runs in the browser. When you load up a new website, you download very thin code that comes with all the necessaries to be able to communicate with a server.
```
sjudson: > you download very thin code

You tend to download a shitton of code, because logic that used to live server side lives client side. It's one of the reasons sites are getting so bloated.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## The complexity of web applications
+
+In the 1990s world of basic web development, you had monolithic applications that handled user management, stored sensitive data, processed requests and returned some HTML to our screen. Those applications acted as both our server and our client application, all rolled into one.
+
+In the years that followed, developers made that HTML prettier and prettier, and more interactive. But they essentially still had one application which kept our users' credentials and session secure, for the most part, in the backend and still returned HTML to the front-end. Our monolithic application is **one single application**, and it's just sending HTML down to our browser, which is more or less just a dumb client that lets us click on our new pretty buttons and links.
+
+At this point, the browser is not actively involved in any logic. All you need to do with each request is to identify the user is who they were at the start of their browsing session. In such a case you would use cookies, because that's what they were designed for: a single party that has no need to expose anything with any potential sensitivity.
```
sjudson: > At this point, the browser is not actively involved in any logic.

Um, yes it is. It, e.g., has to handle cookies and other headers correctly.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+Roll on two decades of user experience improvements and technological advancements and you land smack bang in the world of single-page applications (SPAs). An SPA is essentially software that runs in the browser. When you load up a new website, you download very thin code that comes with all the necessaries to be able to communicate with a server.
+
+The server remains essentially the same as our old 1990s application, but it’s learned some tricks. It has had an upgrade. It only talks in computer language now, HTTP requests sending and receiving JSON payloads. It’s now called an API and it doesn't return HTML anymore. Instead, it leaves the HTML up to the SPA.
+
+Web tokens were designed for use in cases where there are *two or more applications*, such as our API and our SPA. It allows the SPA to talk to the API, where you can identify the authorized user. If you don't protect your tokens with all the responsibility you give other credentials and then the token is accessed by a bad actor, they could access your API on behalf of that user.
```
sjudson: > Web tokens were designed for use in cases where there are *two or more applications*, such as our API and our SPA.

Web tokens were designed to serialize information across security domains. They got shoehorned into this anti-pattern.

sjudson: >  If you don't protect your tokens with all the responsibility you give other credentials and then the token is accessed by a bad actor, they could access your API on behalf of that user.

You need to make clear that this insufficient protection is _inherent_ in this design.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+**Now imagine that the API is a bank, and the SPA is a user's online banking software.**
```
sjudson: I'd make this a less concerning model. Otherwise you'll get a lot of "yeah I'd never do this as a bank, but I just have a messaging app" responses. Do like a photo editing app.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Auth0’s old guides, they store tokens in localStorage!
+
+We know and we’re working really hard to update our existing guides.
+
+A lot of people will still go ahead and use `localStorage` assuming they’ll never be vulnerable to XSS attacks, having taken the appropriate steps to protect against them.
```
sjudson: Again, not the only concerning attack vector.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+That’s okay, but the responsible thing for Auth0 to do **is teach you the most secure way**.
```
sjudson: Not technical, but using the word "teach" is pretty off-putting. We've always been wrong, and aren't really in a position to be "teaching". I'd say "present".

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Your returning users can still re-authenticate
+
+So we’ve discussed the risk of storing tokens with the `Web Storage API` and how you might mitigate the risk if you're determined to do it that way. But as we said, that’s not enough. We need to show you the right way an SPA can authenticate and re-authenticate a returning user without that token appearing from `localStorage`.
+
+The [`auth0-js` SDK](https://auth0.com/docs/libraries/auth0js) `checkSession()` method allows you to acquire a new token from Auth0 for a user who is already authenticated against Auth0 for your domain. The method accepts any valid OAuth 2.0 parameters that would normally be sent to our `authorize()` method, which is used to initiate authentication in the first place. This allows you to re-authenticate returning users without any dramatic re-engineering. Users won't be presented with a login screen unless required. `checkSession` will get a new token transparently when certain conditions are met.
```
sjudson: I'd be explicit that you're exploiting SSO at the Auth0 tenant here. Just as w/ SSO a user goes to a new app and is automatically logged in, so to can hat be used to keep the user status updated locally without asking them to reenter credentials. Also if you're not clear those "certain conditions" are the user having a valid upstream session this looks like more insecure magic.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+The Auth0 guides would normally include an authentication utility script to provide some much-needed functions for authentication on an SPAs. Below you'll find an example using the `auth0-js` SDK. **This example isn't working code and shouldn't be expected to run**, but it gives you a good idea of what methods you should be looking to use in order to avoid storing tokens with the `Web Storage API`.
```
sjudson: Shouldn't this just link to our actual docs? I don't like the idea of non-functional "guidance" code like this, and I'm not going to give it a proper review as such.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+  handleReturningUser();
+});
+```
+
+At this point you might be wondering how `checkSession()` works, right?
+
+`checkSession()` works by creating an invisible iframe which makes a request to `the-users-domain.auth0.com/authorize`. What this request does is return a new `token` to the application.
+
+It returns tokens if the user is still logged in to the authentication server. The authentication server uses normal `web sessions` to know whether the user is authenticated or not, just like like going back to Facebook a day later and it logging users back in. They can still get a `token` from the authentication server while their session still exists.
```
sjudson: > normal `web sessions`

Just say it uses "a session".

sjudson: > They can still get a `token` from the authentication server while their session still exists.

The user doesn't get a token the client does, if the user authenticates and approves it. While they have a session, that can be done invisibly (as they've already both authenticated and approved).

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+That session will expire eventually, according to authorization server configuration. If the session has expired, the application will receive an `interaction_required` error in the response, which is when it’s time to log in again!
```
sjudson: Note that this length of time is configurable with your Auth0 tenant.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Conclusion
+
+Access tokens are to an application what usernames and passwords are to a login box, representing a user until expiry.
```
sjudson: They represent a delegation to the client of a subset of the user's permissions. An `id_token` or response from `/userinfo` (in OIDC) represents a user.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+Unnecessary risks with access tokens such as storing them in `localStorage` and storing them incorrectly in a cookie leaves them vulnerable to many attack vectors. While these attacks can be defended against, it’s only responsible of us to teach you the most secure way to handle authentication.
```
sjudson: Again, reiterate that at a foundational level exposing tokens to the browser is less secure. It's not just XSS which you can protect against. It's a much more exposed piece of software than a web application running on a privately managed server. Other attacks matter too.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+That is why we're recommending that access tokens aren't stored with the `Web Storage API`, and if they're stored in a cookie then you should apply `HttpOnly`, `SameSite` and `Secure`  directives. Additionally, cookies can have `__Secure` and `__Host` prefixes. For more information on securing cookies, [read about common threats in web application security and how to prevent them](https://auth0.com/blog/common-threats-in-web-app-security/).
```
sjudson: I wouldn't event recommend putting them in a cookie. It's better, but not by much. Just have a backend, or forget the token as soon as you use it and employ SSO behavior to get them as needed.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
+---
+layout: post
+title: "Token Security: Stop Putting Your Users At Risk"
+description: "Access tokens are a credential that can be used by a client to access a service, like an API. Like any credential, passwords for instance, they should be kept safe and secure."
```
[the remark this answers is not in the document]
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+- jwt
+- auth
+- breaches
+related:
+- 2018-02-27-8-ways-to-avoid-healthcare-breaches
+- 2018-02-14-what-is-data-security
+- 2018-03-22-cambridge-analytica-and-facebook§t
+---
+
+**TL;DR:** We explore keeping your tokens and your user’s data safe. The [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) has been used as an example of how to store access tokens, especially in single-page applications (SPAs). Auth0 has an alternative to storing a token for a persistent user session, so your returning users can stay logged in without putting them at unnecessary risk.
```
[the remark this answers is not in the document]
lukeoliff: fixed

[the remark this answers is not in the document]
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+
+## Storing tokens on the front-end, don’t do it
+
+Access tokens are a credential that can be used by a client to access a service, like an application programming interface (API). Like any credential, passwords for instance, they should be kept safe and secure.
+
+{% include tweet_quote.html quote_text="Are you putting your users at unnecessary risk? If you store access tokens in your front-end app, you might be!" %}
+
+You should not be storing anything as sensitive as credentials anywhere they can be accessed by a third party. Storing them in your front-end application is asking for trouble. Attackers are actively looking for places you might be exposing sensitive data and they’re using more elaborate methods all the time. The last thing you need to do is make it easy for them.
```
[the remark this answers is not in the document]
lukeoliff: reworded, fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
```diff
@@ -0,0 +1,192 @@
...
+- 2018-02-27-8-ways-to-avoid-healthcare-breaches
+- 2018-02-14-what-is-data-security
+- 2018-03-22-cambridge-analytica-and-facebook§t
+---
+
+**TL;DR:** We explore keeping your tokens and your user’s data safe. The [Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) has been used as an example of how to store access tokens, especially in single-page applications (SPAs). Auth0 has an alternative to storing a token for a persistent user session, so your returning users can stay logged in without putting them at unnecessary risk.
+
+## Storing tokens on the front-end, don’t do it
+
+Access tokens are a credential that can be used by a client to access a service, like an application programming interface (API). Like any credential, passwords for instance, they should be kept safe and secure.
```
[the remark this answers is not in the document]
lukeoliff: fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded, fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded, fixed

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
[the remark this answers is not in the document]
lukeoliff: reworded it

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: Reworded slightly. This entire section was paraphrasing guidance from Jared. I think it works for the target audience.

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: My understanding is that we haven't updated all of our docs or quickstarts yet and that this article is the first step to adjusting our position.

For example https://auth0.com/docs/security/store-tokens we still have this guidance that doesn't say "use SSO, don't store tokens"

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

_posts/2018-03-28-token-security-stop-putting-your-users-at-risk.markdown
lukeoliff: reworded

PR #1479, opened 2018-04-25
What's new in Angular 6?

unicodeveloper:


_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+
+{% highlight html %}
+{% raw %}
+  <ng-template>some template content</ng-template> 
+{% endraw %}
+{% endhighlight %}
+
+## 3. Better URL Serialization
+
+Before now, there were issues around routes and URL serialization such as this below:
```
dan-auth0: Before now, there were issues around routes and URL serialization such as this below: => Before now, there were issues around routes and URL serialization such as this ~below~:

There's no need to add below in this context as the image/example directly follows.

_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+    customElements.define('my-name', NgElementConfig);
+  }
+}
+```
+
+### 5. Forms Validation in Angular 6
+
+Before now, `ngModelChange` was always emitted before the underlying form control was updated. 
+
+If you had a handler for the `ngModelChange` event that checked the value through the control, the old value will be logged instead of the updated value. This is not the case if you pass the value through the `$event` keyword directly.
```
dan-auth0: the old value will be logged instead of the updated value =>  the old value **would** be logged instead of the updated value

_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+
+```js
+onChange(ngModel: NgModel) {
+   console.log(ngModel.value); // logs updated value
+}
+```
+
+## 6. Multiple Validators for Form Builder Array
+
+In previous versions of Angular, you can set only one validator on `FormArray` field with the `FormBuilder.array` method.
```
dan-auth0: In previous versions of Angular, you can set only one validator => In previous versions of Angular, you **could** set only one validator

dan-auth0: one validator on `FormArray` field => one validator on a `FormArray` field

_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+
+In Angular 6, you can set multiple validators for the `FormBuilder.array` method:
```
dan-auth0: In Angular 6, you can set multiple validators for the `FormBuilder.array` method: => In Angular 6, you can set multiple validators **with** the `FormBuilder.array` method:

_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+}
+```
+
+## 7. Token Marking for Runtime Animation Context
+
+In Angular 6, it's now possible to determine which animation context is used for a component at runtime. A token is provided as a marker to determine whether the component is running a `BrowserAnimationsModule` or `NoopAnimationsModule` context at runtime.
+
+## 8. Hello Schematics
+
+**Schematics** is a new scaffolding library that's used by the Angular CLI to generate custom templates. The Angular team has always been keen on improving developer productivity, thus the birth of **schematics**.
```
dan-auth0: , thus the birth of **schematics**. => , which explains the birth of **schematics**.

I think this phrasing makes the sentence flow better.

_posts/2018-04-24-whats-new-in-angular6.markdown
```diff
@@ -0,0 +1,284 @@
...
+schematics ng-lib-schematics:lib-standalone --name avengers
+```
+
+A new `lib` directory will be generated for you inside the `src` folder. The `lib` directory ships with a sample demo and the build tools necessary for a typical Angular package.
+
+Check out this [deep and excellent guide to Schematics](https://blog.angular.io/schematics-an-introduction-dc1dfbc2a2b2).
+
+## Deprecations and Other Updates
+
+* Angular 6 ships with Rxjs 6.0.0.
```
dan-auth0: Let's include a link to the RxJS repo here: https://github.com/ReactiveX/rxjs

Angular 6 ships with [Rxjs 6.0.0](https://github.com/ReactiveX/rxjs).

dan-auth0:
Thanks for the article and integrating the feedback, @unicodeveloper!

PR #1493, opened 2018-05-03
Ruby on Rails—Killer Workflow with Docker (Part 1)

brunokrebs:


_posts/2018-05-15-ruby-on-rails-killer-workflow-with-docker-part-1.md
```diff
@@ -0,0 +1,658 @@
...
+docker-compose up --build
+```
+
+You will see that `step 1/nn: FROM starefossen/ruby-node:2-8-stretch` is available **instantly** as a layer (that is if you have already tried the initial `docker run` with a single line of `Dockerfile`).
+
+That's Docker reusing the Ruby image downloaded earlier. The whole process that ran last time is saved for you. So it is only from the second line of code within the `Dockerfile` that matters now.
+
+You'll see all the steps within the `Dockerfile` executed dutifully during the build process. Each of them creates a **layer** that you can recognize by their hash, that looks like `1adb3ee3e245`. 
+
+Once build steps are over, you'll see `db` service being started first. And then the `app` service starts, but **stops** when trying to run `bundle exec rails s`, as we do not have `rails` gem within the environment yet.
```
vijayabharathib: I was wrong here. app service fails NOT because "we don't have `rails` gem", we do have it in the container after `docker-compose up --build`, but because "as you have not created a Rails project yet".

_posts/2018-05-15-ruby-on-rails-killer-workflow-with-docker-part-1.md
```diff
@@ -0,0 +1,658 @@
+---
+layout: post
+title: "Ruby on Rails—Killer Workflow with Docker (Part 1)"
+description: "Learn how to set up a killer dockerized workflow that will raise your productivity while developing Ruby on Rails applications."
+longdescription: "In this series, you will learn how to set up a killer dockerized workflow that will raise your productivity while developing Ruby on Rails application. You will use tools like Docker, Docker Compose, Travis, and Heroku to set up a state-of-the-art workflow."
+date: 2018-05-15 08:30
+category: Technical Guide, Backend, Ruby On Rails
+design:
+  bg_color: "#333333"
+  image: https://cdn.auth0.com/blog/logos/node.png
```
vijayabharathib: This one shows a NodeJS image for the post (when I launched local server on guest-writer). May be a Rails/Docker image would be appropriate?

_posts/2018-05-15-ruby-on-rails-killer-workflow-with-docker-part-1.md
```diff
@@ -0,0 +1,658 @@
...
+- ruby
+- rails
+- ci
+- cd
+- ci/cd
+- continuous-integration
+- continuous-delivery
+- auth0
+related:
+- 2018-05-17-ruby-on-rails-killer-workflow-with-docker-part-1
```
vijayabharathib: You might have meant to add a link to part 2 (this seems to be recursive). Also, file name for part 1 starts with **2018-05-15**, here the date is **2018-05-17**.

_posts/2018-05-15-ruby-on-rails-killer-workflow-with-docker-part-1.md
```diff
@@ -0,0 +1,658 @@
...
+  gem 'guard-minitest', '~>2.4.6', require: false
+  gem 'rack-livereload'
+  gem 'foreman'
+end 
+```
+
+Ensure you **do not remove** any existing gems in the process, just add this group right at the end of the `Gemfile`. Remember the usual drill to build the image to include these new gems. 
+
+```bash
+docker-compose up --build
```
vijayabharathib: worth adding a `docker-compose down` before `docker-compose up --build`. When I created the project again, I left the old container running in another terminal and the rebuild wasn't successful. Bringing container down might help.

PR #1590, opened 2018-07-05
SEO - Adding metadescription field and reworking post includes - WIP

scottmathson-auth0:
Work in progress, more to come on this. Starting point to adding a post subtitle (short, front-end description) and separating `description` into being the richer, longer SERPs description.


_includes/amp/banner_post.html
```diff
@@ -17,7 +17,12 @@
 
   {% include amp/entry_image.html %}
   <h1 class="post-title" itemprop="headline">{{current.title}}</h1>
-  {% if current.description %}
```
kmaida: I'd recommend leaving `description` as the name of the field that shows visibly in the front end rather than changing its purpose and adding a new field for the UI. This will lower the chances of people misusing it. If people are used to putting the subtitle in a `description` field, switching to use an identically-named field for a different purpose is higher barrier to entry and a higher chance that it won't be used correctly.

I'd recommend leaving `description` (for UI) and adding `searchdescription` (or `metadescription`, or something similar) for SEO.

scottmathson-auth0:
Merging, but noting that AMP still needs title/description separation work.

scottmathson-auth0:
Keep branch alive please for further work to be done.

PR #1661, opened 2018-08-13
Create profile_card.html include.

dan-auth0:
This file is needed for Auziro profiles in Tech@Auth0 blog posts.


dan-auth0:
Merging as temporary fix while waiting on styl change process.

PR #1682, opened 2018-08-21
Migrate Symfony from auth0.js

alexisluque:


kmaida:
@lukeoliff This post has been updated to not use auth0.js on the client side, but rather, implement the flow for traditional web apps. Can you please give the changes a quick read and merge if all looks good? Thanks!

PR #1702, opened 2018-08-27
Add post_length to Rakefile and post length bash script

samjulien:
This commit adds the `post_length` attribute to the Rakefile and adds a bash script called `postlength.sh` to the root of the repo. This script generates a score based on the word count of the file passed into the script. Authors can run this script and add that score to post front matter as `post_length`.


kmaida:
Add the notes to the cheatsheet markdown file too, and then merge. 👍

PR #1710, opened 2018-08-29
Add Japanese quotes to Tweet Quote include.

dan-auth0:


kmaida:
Someone had mentioned that Twitter wasn't using the `-jp-` version of the post. Is this fixed here? The code looks OK to me. Can you confirm, @dan-auth0?

dan-auth0:
@kmaida Yup! It was because the test file did not have the `jp-` prefix. We tested it with live posts and the correct URL is already in place. 

The only outstanding item is the translation of `via`. I am awaiting a response from our teammate. I think we could still merge this one in the meantime. What do you think?

kmaida:
Confirmed that "via" does not need to be translated and should be removed.

PR #1724, opened 2018-08-31
Add NgRx authentication article and update cheatsheet

samjulien:


dan-auth0:
@samjulien Overall, I recommend changing the `—` to `&mdash;` throughout the article.

_posts/2018-09-01-add-auth-to-ngrx.markdown
```diff
@@ -0,0 +1,1081 @@
...
+```
+
+Let's break down what's happening in each of these.
+
+- Login — calls the `login` method on `AuthService`, which triggers Auth0. Does not dispatch an action.
+- Login Complete — calls `handleLoginCallback` on `AuthService`, which returns an observable of the parsed hash result. If there's a token, this effect calls `setSession` and then dispatches the `LoginSuccess` action. If there's not a token, the effect dispatches the `LoginFailure` action with the error as its payload.
+- Login Redirect — This effect happens when `LoginSuccess` is dispatched. It redirects the user to `home` and does not dispatch a new action.
+- Login Failure — This effect happens when `LoginFailure` is dispatched. It redirects the user back to `login` and does not dispatch a new action.
+
+If you run the application with `ng serve`, you should now be able to successfully log in to your application using Auth0! You'll see a login screen similar to this:
```
dan-auth0: The log in is not working there. I was able to catch the following error:
`error: "access_denied", errorDescription: "Service not found: http://localhost:3001",`
The error is gone and the auth flow is successful when `audience: environment.auth.audience,` is commented out from `src/app/auth/services/auth.service.ts`

_posts/2018-09-01-add-auth-to-ngrx.markdown
```diff
@@ -0,0 +1,1081 @@
...
+
+<p style="text-align: center;">
+  <img src="https://cdn.auth0.com/blog/add-auth-to-ngrx/auth0-book-login.png"
+  style="max-width:400px" alt="Our Auth0 login screen works!">
+</p>
+
+You'll then be redirected to the `home` route, where you can click the button to view the book collection. Of course, we can't log out yet, so let's add the effects for that now.
```
dan-auth0: I suggest mentioning that they need to create a new user or log in with an existing one. Linking to how to create users in the dashboard would be good too: https://auth0.com/docs/dashboard/creating-users-in-the-management-portal

_posts/2018-09-01-add-auth-to-ngrx.markdown
```diff
@@ -0,0 +1,1081 @@
...
+
+## Review and Where to Go From Here
+Congratulations on making it to the end! We've covered a lot of ground here, like:
+
+- Using @ngrx/schematics to quickly generate new code
+- Defining global authentication state
+- Using selectors to access authentication state
+- Setting up Auth0 to handle your authentication
+- Using effects to handle the login and logout flows
+
```
dan-auth0: There's one issue: if the user refreshes the app, the authentication state is not being persisted. For this, we need to create a flag in local storage. The flow is present on the `AuthService` in https://auth0.com/blog/how-to-create-angular-apps-with-auth0-and-stackblitz/

_posts/2018-09-01-add-auth-to-ngrx.markdown
```diff
@@ -0,0 +1,1081 @@
...
+We first need to import the `auth0-js` library, our environment configuration, and two things from RxJS:
+
+```typescript
+import * as auth0 from 'auth0-js';
+import { environment } from '../../../environments/environment';
+import { Observable, of } from 'rxjs';
+```
+
+Next, we need to set some public properties on our class. We'll need an Auth0 configuration property, properties to store our token and expiration date, and a boolean property to set whether the user is authenticated. Add these before the class `constructor`:
```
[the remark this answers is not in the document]
samjulien: Agree to disagree here -- I've always heard/said "set properties on a class" because they are public properties on the class itself.

_posts/2018-09-01-add-auth-to-ngrx.markdown
samjulien: I've removed the references to `audience` and added a note about why we're not using it.

PR #1746, opened 2018-09-07
Updating T-Mobile post

scottmathson-auth0:


scottmathson-auth0:
@RamiroND FYI, no need for `metatitle` if they're exact same/duplicates. Also making some other edits.

PR #1749, opened 2018-09-10
Auth0 Announces New VP of Sales, Americas

RamiroND:


scottmathson-auth0:
FYI @RamiroND `Metadescription` capitalized does not recognize and honor that field in `<meta name="description"`...

PR #1757, opened 2018-09-11
Re-prioritizing title over metatitle for social sharing

scottmathson-auth0:
Keeping post field `metatitle` solely for the value it brings to SERPs, yet changing `_includes` to not source this as what’s pulled into sharing on social as per conversation with @RamiroND.

That will remain/revert to pulling the `<h1>` (`title` field). Think this will be best in the long run and lessens opportunities for disconnect between titles.


scottmathson-auth0:
Reverts some work done here https://github.com/auth0/blog/pull/1590

